PR #3, opened 2019-08-25
Create fa_clone_vol_to_remote_array.yaml

dnix101:
SUMMARY
Example playbook to copy a volume from one FlashArray (source) to another (target), using a protection group to create and replicate a snapshot to the remote array. Then creating a volume clone on the remote array. Finally the snapshots on both the source and target are cleaned up.

ISSUE TYPE
New Module Pull Request

COMPONENT NAME
fa_clone_vol_to_remote_array.yaml


dnix101:
Understood 😉

PR #30, opened 2019-12-18
Add support for VMware protocol-endpoints and a few fixes

sdodsley:
##### SUMMARY
Add support for VMware protocol endpoints.
Fix some logic errors
Add recover volume from eradicated state capability

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME
purefa_volume
purefa_endpoint

##### ADDITIONAL INFORMATION
`purefa_volume` module modified to ensure volume actions are not taken on an endpoint with a corresponding name and also fix a poor `mutually_exclusive` parameter

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/README.md
+++ b/collections/ansible_collections/purestorage/flasharray/README.md
@@ -11,6 +11,7 @@ The Pure Storage FlashArray collection consists of the latest versions of the Fl
 - purefa_dns - manage the DNS settings of the FlashArray
 - purefa_ds - manage the Directory Services of the FlashArray
 - purefa_dsrole - manage the Directory Service Roles of the FlashArray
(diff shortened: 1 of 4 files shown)


sdodsley:
@dnix101 fixed

dnix101:
@sdodsley Thanks, no errors anymore. However delete and rename procedures fail. 

Deleting a just created endpoint returns:
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "Delete endpoint testendpoint1 failed."}`
Although I am able to delete it from the GUI.

Renaming the just created endpoint returns:
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "Target endpoint foo-bar already exists."}`
Even though foo-bar doesn't exist (either as volume or endpoint). I also tried an other name.

PR #32, opened 2019-12-23
Update purefa_hg to enable LUN ID is only one volume defined for HG

sdodsley:
##### SUMMARY
If only one volume is provided allow a LUN ID to be specified for that volume

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_hg

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_hg.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_hg.py
@@ -40,6 +40,13 @@ options:
     type: list
     description:
     - List of existing volumes to add to hostgroup.
+  lun:
+    description:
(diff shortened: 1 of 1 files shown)


sdodsley:
@dnix101 works for me.... Your syntax looks correct. The volume has to exist. This module doesn't create the volume (unlike purefa_host)

dnix101:
@sdodsley Just tested some more. It works for a normal volume (name John), but not for a volume in a Pod (pod::John). The volume in the pod is connected, but the lun ID is not set.

sdodsley:
@dnix101 so I tested with a pod volume and that connects fine with a LUN ID. How can a volume be connected without a LUN ID? If the volume is already connected this module will not connect it to something else. Can you perform the action you want in the module through the GUI? If yes, then I need more details for this corner case to code for it

PR #40, opened 2020-01-10
Add support for network interface configuration

sdodsley:
##### SUMMARY
Add new module to reconfigure physical and virtual network interfaces
Update README with new module

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_network.py

##### ADDITIONAL INFORMATION
You can delete an interfaces IP address by specifying the IP address as `0.0.0.0/0`. 
Removing an IP address is supported for ports with the `management` service configured. This is to stop the array being inadvertently being bricked.

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
@@ -1,0 +1,231 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2020, Simon Dodsley (simon@purestorage.com)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: purefa_network
+short_description:  Manage network interfaces in a Pure Storage FlashArray
+description:
+    - This module manages the physical and virtual network interfaces on a Pure Storage FlashArray.
+    - To manage VLAN interfaces use the I(purefa_vlan) module.
+    - To manage network subnets use the I(purefa_subnet) module.
+    - To remove an IP address from a non-management port use 0.0.0.0/0
+author: Pure Storage Ansible Team (@sdodsley) <pure-ansible-team@purestorage.com>
+options:
+  name:
+    description:
+      - Interface name (physical or virtual).
+    required: true
+    type: str
+  state:
+    description:
+      - State of existing interface (on/off).
+    required: false
+    default: present
+    choices: [ "present", "absent" ]
+    type: str
+  address:
+    description:
+      - IPv4 or IPv6 address of interface in CIDR notation.
+      - For example IPv4: ddd.ddd.ddd.ddd/dd
+      - For example IPv6: xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx/xxx
+      - To remove an IP address from a non-management port use 0.0.0.0/0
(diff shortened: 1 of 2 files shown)


sdodsley:
@dnix101 History :) When I first wrote that module many months ago I did the split, but when I did VLAN I forgot I had split them.
It makes more sense now to keep them together. I'll rework the module.
Also, I'll look at deleting the IP address, but that could cause the array to be bricked. Maybe I will only allow IP deletion for non-managment ports

dnix101:
That makes sense, looking forward to the next reworked version. I also agree removing an IP should be limited to non management interfaces indeed.

sdodsley:
@dnix101 New version now up with controller information included in the `name` parameter and the ability to remove an IP address from non-management and non-app ports (eg `vir0` and `@offload`).

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
```diff
@@ -0,0 +1,240 @@
...
+
+    module = AnsibleModule(argument_spec,
+                           supports_check_mode=True)
+
+    if not HAS_NETADDR:
+        module.fail_json(msg='netaddr module is required')
+
+    state = module.params['state']
+    if '.' in module.params['name'] or '.' in module.params['controller']:
+        module.fail_json(msg='Incorrect controller or interface provided.')
```
dnix101: Lines 223 and 224 should be removed, since controller is no longer a valid param and name can include a '.' now.

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
```diff
@@ -0,0 +1,240 @@
...
+    array = get_system(module)
+    interface = _get_interface(module, array)
+    if not interface:
+        module.fail_json(msg="Invalid network interface specified.")
+
+    if state == 'present':
+        update_interface(module, array, interface)
+    elif state == 'absent':
+        disable_interface(module, array)
```
dnix101: If you are trying to set an IP address (or remove it) and combine that with `state: absent`, the IP address is not set. Seems to be as designed, but I'm not sure if it was ment to be this way.

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
```diff
@@ -0,0 +1,240 @@
...
+    fa_url: 10.10.10.2
+    api_token: c6033033-fe69-2515-a9e8-966bb7fe4b40
+
+- name: Enable virtual network interface vir0
+  purefa_network:
+    name: vir0
+    state: present
+    fa_url: 10.10.10.2
+    api_token: c6033033-fe69-2515-a9e8-966bb7fe4b40
+'''
```
dnix101: Add example to remove IP address info:
```
- name: Remove IP address from iSCSI interface ct0.eth4
  purefa_network:
    name: ct0.eth4
    address: 0.0.0.0/0
    gateway: 0.0.0.0
    fa_url: 10.10.10.2
    api_token: c6033033-fe69-2515-a9e8-966bb7fe4b40
```

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
```diff
@@ -0,0 +1,240 @@
...
+extends_documentation_fragment:
+    - purestorage.fa
+notes:
+    - Requires the netaddr Python package on the host.
+requirements:
+    - netaddr
+'''
+
+EXAMPLES = '''
+- name: Configure and enable network interface ct0.eth8
```
dnix101: Change ct0.eth8 to ct0.eth0

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
sdodsley: Yes - I'll rewrite this so you can enable/disable and change settings at the same time.

PR #43, opened 2020-01-15
Add support deletion of snapshots from an offload target protection group

sdodsley:
##### SUMMARY
Add feature to be able to delete snapshots on offload target protection groups.
Also, fix a stylist error with a variable name

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_pgsnap.py
purefa_snap.py

##### ADDITIONAL INFORMATION
Use the `on` parameter to specify the offload target name and ensure that the snapshot name
is of the format `<arrayname>:<pgname>` and a correct `suffix` is provided

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pgsnap.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pgsnap.py
@@ -61,6 +61,11 @@ options:
     - If not supplied this will default to the volume defined in I(restore)
     type: str
     version_added: 2.8
+  offload:
+    description:
+    - Name of offload target on which the snapshot exists.
+    - This is only applicable for deletion and erasure of snapshots
+    type: str
   now:
     description: Whether to initiate a snapshot of the protection group immeadiately
     type: bool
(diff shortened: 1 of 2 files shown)


sdodsley:
@WormNRifle These modules are designed to work with Ansible collections. If you are using regular Ansible and replacing the upstream modules then this will happen.

WormNRifle:
Understood. 

I will look into that.

Thank you.

WormNRifle:
I pulled the project down via collections, swapped out the purefa_pgsnap.py file in the collection and reran my playbook.

I'm getting the following:

> "Unsupported parameters for (purefa_pgsnap) module: True Supported parameters include: api_token, apply_retention, eradicate, fa_url, name, now, overwrite, remote, restore, state, suffix, target"

sdodsley:
@WormNRifle Ah - there is a typo in the examples `on` should be `offload`. I have just pushed a fix.

WormNRifle:
So, can I assume the OFFLOAD_API check is the version the offload code that is installed on my pure array?

sdodsley:
It is checking to ensure the version of Purity running on the array is high enough to support the Offload functionality.

WormNRifle:
Thank you for that verification.

PR #75, opened 2020-04-20
Ensure hosts are lowercase to stop case inconsistency

sdodsley:
##### SUMMARY
Ensure all hostnames are lowercase to stop errors with pre-created FA hosts having a mismatch with the true host name.

Example: 
host = Linux
array object = linux

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_host.py
purefa_hg.py
purefa_pg.py

#### ADDITIONAL INFORMATION
Fix a small typo and bump the Collection version number

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/galaxy.yml
+++ b/collections/ansible_collections/purestorage/flasharray/galaxy.yml
@@ -1,6 +1,6 @@
 namespace: purestorage
 name: flasharray
-version: 1.2.2
+version: 1.2.3
 readme: README.md
 authors:
 - Pure Storage Ansible Team (@sdodsley) <pure-ansible-team@purestorage.com>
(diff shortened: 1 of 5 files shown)


pbouffetps:
With this modification, my playbook don't work anymore because the plugin convert to lowercase the name of the host.
For example:
with
     purefa_host:
        host: LAB-esxi01
        preferred_array:
          - array3
        fa_url: "{{ tgt_fa_url }}"
        api_token: "{{ tgt_fa_api_token }}"

I get this error:
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Host lab-esxi01 creation failed."}

Moreover the convention is defined as this in the FA User Guide:
Object Names
Valid characters are letters (A-Z and a-z), digits (0-9), and the hyphen (-) character. The first and last
characters of the name must be alphanumeric, and the name must contain at least one letter or '-'.

So howto handle this?

sdodsley:
@pbouffetps Whilst the user guide states that uppercase characters are allowed, normal Linux rules are that all hostnames should be lowercase and we tend to pull hostnames in dynamic playbooks from the ansible hostname fact which will be lowercase.
On an FA there is no differentiation between host 'U1' and host 'u1' so we will have to look into the logic a little more to come up with a solution.
Please raise an Issue for us to look into this some more.

PR #107, opened 2020-08-24
Update case-sensitivity code

sdodsley:
##### SUMMARY
Update case-sensitivity code
FlashArray hostname/protection group name/hostgroup name rules state that upper case is allowed
Add documentation notes that names are case sensitive but with a note that name uniqueness is NOT case sensitive.
For example: 
    * Allowed hostnames - hosta and hostA
    * However hosta = hostA in a FlashArray

Ensure host that is in a hostgroup can be correctly deleted

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_host
purefa_hg
purefa_pg


dnix101 (approved):
lgtm

PR #110, opened 2020-09-30
Add support for API Client management

sdodsley:
##### SUMMARY
New module to support API Client management

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_apiclient.py

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_apiclient.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_apiclient.py
@@ -1,0 +1,215 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2020, Simon Dodsley (simon@purestorage.com)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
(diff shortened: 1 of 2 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_apiclient.py
```diff
@@ -0,0 +1,215 @@
...
+    - Define whether the volume should exist or not.
+    default: present
+    choices: [ absent, present ]
+    type: str
+  role:
+    description:
+    - The maximum role allowed for ID Tokens issued by this API client
+    type: str
+    choices: [readonly, ops_admin, storage_admin, array_admin]
+  issuer:
```
dnix101, line 40: The role is only accepted during creation and ignored if the API client already exists. Is that intentional? Can also not be changed using the GUI for an existing API client btw. However I would expect an error or warning?

rdeenik (dismissed):
some comments on documentation, functionality lgtm

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_apiclient.py
sdodsley: Yes - you cannot change an API client once it is created. All you can do is enable/disable the client

PR #111, opened 2020-09-30
Add support for FA Files filesystems

sdodsley:
##### SUMMARY
New module to create, delete, eradicate, rename and recover filesystems.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_fs.py


dnix101 (commented):
If you delete and then delete with eradicate, the fs is not eradicated. Is that by design?

sdodsley:
Yes - no trash can. Also you can only delete if there are no active exports connected or there is content in the directory - there is no way to test for the latter so I try and delete and if it fails I use the error message the API call returns for clarity.
Still working this module though - some updates coming later today

dnix101:
> Yes - no trash can. Also you can only delete if there are no active exports connected or there is content in the directory - there is no way to test for the latter so I try and delete and if it fails I use the error message the API call returns for clarity.
> Still working this module though - some updates coming later today

Actually discard my comment. Was an error in my play book. Creation, rename, delete and then delete and eradicate works fine.

dnix101 (approved):
shipit

PR #120, opened 2020-10-28
Fix QoS for low Purity arrays, AC and checkmode issue

sdodsley:
##### SUMMARY
Fix issue where check_mode fails due to missing variable
Fix issue when trying to define IOPs QoS on an array with incorrect Purity version
Do not allow QoS to be set on any volumes in AC pods unless at correct Purity version

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_volume.py

##### ADDITONAL INFORMATION
The limit on no QoS for AC volumes was relaxed at Purity//FA 5.2 (REST 1.16)


dnix101 (approved):
shipit

PR #121, opened 2020-11-05
Add multi-volume creation support (Purity 6.0+)

sdodsley:
##### SUMMARY
From Purity 6 we can now create multi-volumes with a single call.

Required information is the number of volumes to create, the index start point and the index character size. Name suffix after index is also supported.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_volume.py

##### ADDITIONAL INFORMATION
New parameters:
 * **count** - Number of volumes to create (Required)
 * **start** - Start number for volume index (Default = 0)
 * **suffix** - String to append to volume name and index

The volumes created will have the name format: _name#suffix_ where # is the volume index number

**!! NOTE !!**
 * Only volume creation is supported by multi-volume at this time.
 * If a multi-volume created volume is deleted or eradicated from the sequence he idempotency of the creation task will fail

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/121_add_multi_volume_creation.yaml
+++ b/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/121_add_multi_volume_creation.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - purefa_volume - Add support for multi-volume creation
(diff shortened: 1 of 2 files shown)


dnix101 (changes requested):
Two comments:
- Creation of multiple volumes in a POD or VG returns errors
- It would be nice if the module would return the name, size, serial, etc of the volumes as output, so that you can `register` the output and use that in the rest of the playbook. When creating a single vol this information is being returned by the module.

sdodsley:
@dnix101 Let me check the pod and VG errors.
As to the output - unfortunately, the REST response is not in a form that enables us to match volume with size and serial number

dnix101:
> @dnix101 Let me check the pod and VG errors.
> As to the output - unfortunately, the REST response is not in a form that enables us to match volume with size and serial number

Any chance we can at least return the names?

sdodsley:
Not really - it all just comes out as a string of values with no correlation between them. I'll see if there is anything I can produce

sdodsley:
@dnix101 Pods and VGs now work and you also get the same return from multi-volume creation as you do from a single volume

dnix101 (approved):
Perfect, shipit!

PR #122, opened 2020-11-05
Add multi-host creation support (Purity 6.0+)

sdodsley:
##### SUMMARY
From Purity 6 we can now create multi-hosts with a single call.

Required information is the number of hosts to create, the index start point and the index character size. Name suffix after index is also supported.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_host.py

##### ADDITIONAL INFORMATION
New parameters:
 * **count** - Number of hosts to create (Required)
 * **start** - Start number for host index (Default = 0)
 * **suffix** - String to append to host name and index

Optional parameters:
 * **personality** is the only supported optional parameter

The host created will have the name format: _name#suffix_ where # is the host index number

**!! NOTE !!**
 * Only host creation is supported by multi-host at this time.
 * If a multi-host created host is deleted from the sequence the idempotency of the creation task will fail

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
@@ -28,6 +28,10 @@ options:
     - The name of the host.
     - Note that hostnames are case-sensitive however FlashArray hostnames are unique
       and ignore case - you cannot have I(hosta) and I(hostA)
+    - Multi-host support available from Purity//FA 6.0.0
+      B(***NOTE***) Manual deletion of individual hosts created
+      using multi-host will cause idempotency to fail
+    - Multi-host support only exists for host creation
     type: str
     required: true
     aliases: [ host ]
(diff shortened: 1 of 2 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
```diff
@@ -627,49 +715,70 @@ def main():
...
-                    module.fail_json(msg='No target arrays connected to source array - preferred arrays not possible.')
-                else:
-                    current_arrays = [array.get()['array_name']]
-                    api_version = array._list_available_rest_versions()
-                    if NVME_API_VERSION in api_version:
-                        for current_array in range(0, len(all_connected_arrays)):
-                            if all_connected_arrays[current_array]['type'] == "sync-replication":
-                                current_arrays.append(all_connected_arrays[current_array]['array_name'])
+    if module.params['count']:
+        if not HAS_PURESTORAGE:
```
dnix101, line 719: This `if` statement should be moved outside of the `if module.params['count']` context.

dnix101 (changes requested):
New functionality works fine, but think the pypureclient check needs to be moved

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
sdodsley: @dnix101 Why? We only care about the pypureclient being there is `count` is selected as that parameter is only supported in REST 2.2

dnix101: Ah I misunderstood. I'll approve the PR

dnix101 (approved):
shipit

PR #134, opened 2021-01-07
Add support for PGs in Pods

sdodsley:
##### SUMMARY
Missing support for protection groups in AC pods.
If you create a PG in an AC pod currently it will crash the module.
This resolves the bug.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pg.py


dnix101 (approved):
shipit

PR #137, opened 2021-01-12
Add/Update regex check for pgsnap and snap suffix name

sdodsley:
##### SUMMARY
pgsnap module not doing a regex check on the suffix parameter to ensure that the provided string conforms to the 
naming convention rules.
Add regex check to code
Update volume snapshot suffix regex pattern to match pgsnap pattern

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py
purefa_snap.py

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/137_pgsnap_regex.yaml
+++ b/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/137_pgsnap_regex.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+ - purefa_pgsnap - Add check to ensure suffix name meets naming conventions
+ - purefa_snap - Update suffix regex pattern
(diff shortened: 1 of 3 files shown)


sile16:
This is from purity, "Snapshot suffix must be between 1 and 63 characters (alphanumeric and '-') in length and begin and end with a letter or number. The suffix must include at least one letter or '-'."

So maybe a regex of:
^[a-zA-Z]([a-zA-Z0-9-]{0,63}[a-zA-Z0-9])?$"

sile16:
@sdodsley 
Okay, but a simple example of a single digit as a suffix passes your regex but is not allowed by purity, because "The suffix must include at least one letter or '-'."

image

sdodsley:
@sile16 - i used your regex pattern now - let me know if that works better

sile16 (dismissed):

PR #139, opened 2021-01-17
Add support volume restore for volumes in AC based PGs

sdodsley:
##### SUMMARY
Ensure volume snapshots in AC PGs can be recovered.
``target`` parameter can be a local volume name, or a volume in a pod, eg ``volume2``, or ``pod::volume2``.
If you want the restore to go to a volume in the same AC pod, the od name must still be specified.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py


dnix101 (approved):
shipit

PR #140, opened 2021-01-17
Case checking for pod names, including filter down to hgroups

sdodsley:
##### SUMMARY
Enforce all pod names to be lowercase, as there is no case sensitivity in Purity.
If an uppercase name has been manually created, ensure we convert to lowercase when doing checks
adding pod-based volumes to hostgroups
 
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pod.py
purefa_hg.py

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/140_pod_case.yaml
+++ b/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/140_pod_case.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+ - purefa_pod - Ensure all pod names are lowercase for consistency
+ - purefa_hg - Ensure all hostname chacks are lowercase for consistency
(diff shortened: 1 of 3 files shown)


dnix101:
Not sure I understand why the POD name needs to be lower case? FlashArray is indeed case insensitive in the names, but does allow case for objects created. If it's just because the `.difference()` function is case sensitive, we could also use something like this:

```
current_vols = [vol['vol'].lower() for vol in volumes]
new_volumes = []
for vol in module.params['volume']
    if vol.lower() in current_vols: new_volumes.append(vol)
```

That would preserve case, while doing a case-insensitive comparison.

sdodsley:
We force lowercase on volume and host names so for consistency we should do the same for pods, but I like your solution. 
I might use it in other modules as well.

dnix101 (approved):
Approving this PR, but think it would be wise to reconsider forcing lower case. That however would involve multiple modules as mentioned by @sdodsley, so should be in a separate PR.

PR #141, opened 2021-01-26
Add remote volume snapshot support (and snapshot rename)

sdodsley:
##### SUMMARY
Add support for Remote Volume snapshots.
This is new to Purity//FA 6.1 and allows an individual volume to have snapshots send to either a connected array
or a connected offload target.
All appropriate Purity rules are covered for these remote volume snapshots, including the ability to delete snapshots
in use by replication.
New parameters added are:
`offload`
`ignore_repl`

Rename snapshot is also now supported

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_snap.py


dnix101 (approved):
lgtm

PR #166, opened 2021-04-05
Update Remote Assist to support Purity//FA 6

sdodsley:
##### SUMMARY
Fix module to cater for new `connected` value in Purity 6

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_ra.py

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_ra.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_ra.py
@@ -65,16 +65,18 @@ from ansible_collections.purestorage.flasharray.plugins.module_utils.purefa impo
 
 def enable_ra(module, array):
     """Enable Remote Assist"""
-    changed = True
+    changed = False
     ra_facts = {}
-    if not module.check_mode:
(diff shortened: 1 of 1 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_ra.py
```diff
@@ -87,7 +89,7 @@ def disable_ra(module, array):
     """Disable Remote Assist"""
     changed = True
     if not module.check_mode:
-        if array.get_remote_assist_status()["status"] == "enabled":
+        if not array.get_remote_assist_status()["status"] in ["connected", "enabled"]:
```
rdeenik, line 92: Remove `not` from `if not array.get_remote_assist_status`, since currently this will only try to disconnect if it's already disconnected. Otherwise ok.

rdeenik (changes requested):

rdeenik (approved):
lgtm, however just noticed that disabled RA is not idempotent, while enable RA is.

rdeenik (approved):
Shipit

PR #169, opened 2021-04-06
New module for SSL Certificates

sdodsley:
##### SUMMARY
New module to manage SSL Certificates
Options are to create, delete, import or export certificates

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_certs.py

##### ADDITIONAL INFORMATION
management SSL certificate cannot be deleted


rdeenik (approved):
LGTM. We should we add `pycountry` as dependency to https://github.com/Pure-Storage-Ansible/FlashArray-Collection/blob/master/collections/ansible_collections/purestorage/flasharray/requirements.txt

PR #170, opened 2021-04-06
Add new failure state if restoring to stretched pod

sdodsley:
##### SUMMARY
If `target` is in a stretched ActiveCluster pod then fail with "nice" message

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py


rdeenik (approved):
shipit

PR #174, opened 2021-04-07
Allow gateway to be None for network port configuration

sdodsley:
##### SUMMARY
iSCSI ports sometimes do not require a gateway address. 
Ensure that if no gateway address is specified then this is accepted as valid.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_network.py


rdeenik (approved):
shipit

PR #179, opened 2021-04-14
Do not check for stretched target pod if is source = target

sdodsley:
##### SUMMARY
Ensure that we only check for the target pod being stretched if it is not the same as the source pod.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py


rdeenik (commented):
lgtm

rdeenik (approved):
lgtm

Issue #192, opened 2021-04-29
Creating a host with a duplicate wwn. Host is created without wwn.


BenGaba:
**Is your feature request related to a problem? Please describe.**
When creating a new host with Purefa_host with duplicates WWNs ( associated with an other host), the host is still created without WWNs.

**Describe the solution you'd like**
Doing a full rollback. Do not create the host.

**Describe alternatives you've considered**

**Additional context**

sdodsley:
Can you please supply the following information please to assist with resolving this:

FlashArray Purity Version:
Ansible Version:
Ansible FlashArray Collections Version:

BenGaba:
Hello,
Please find the information below :  

FlashArray Purity Version: 5.3.15
Ansible Version: 2.10.7
Ansible FlashArray Collections Version: 1.6.1

sdodsley:
We are testing #193 to resolve this bug.
This will be included in the next Collections release. In the interim, you can use this PR to get the updated code.
Additionally, we suggest you upgrade your FlashArray Collections to the latest version using the command
`ansible-galaxy collection install purestorage.flasharray --force`

BenGaba:
Thank you

sdodsley:
Fixed in #193

PR #193, opened 2021-04-29
Rollback host creation on initiator duplication

sdodsley:
##### SUMMARY
When creating a host setting the FC. iSCSI or NVMe initiators will fail if these are already in use by another host.
Delete the host if this occurs so as not to leave incorrectly configured host objects on the array
Addresses issue #192 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_host.py


rdeenik (approved):
lgtm

PR #199, opened 2021-05-07
Add support for FC port enable/disable

sdodsley:
##### SUMMARY
Add support for enabling/disabling FC ports.
Only available from REST 2.4 or higher

Also, whilst in here, make the whole modules chek_mode functionality more intelligent

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_network.py


rdeenik (commented):
Waiting for release of SDK that support Purity API v2.6 to be able to test.

rdeenik (approved):
shipit

For any one using a older version of py-pure-client, run the following to upgrade:
```pip3 install py-pure-client --upgrade --upgrade-strategy eager```

PR #207, opened 2021-06-17
Fix disable remote assist (purefa_ra)

rdeenik:
##### SUMMARY
Disabling remote assist doesn't seem to work. It tries to disable when already disabled and doesn't do anything when connected.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Related to purefa_ra module.

##### ADDITIONAL INFORMATION


sdodsley (approved):
Good catch

PR #211, opened 2021-07-12
Allow all current host inititators to be removed

sdodsley:
##### SUMMARY
Ensure that setting an empty string for `wwns`, `nqn` or `iqn` will remove all existing initiator ports from a host object.
Ensure `protocol` value is case-insensitive.
Remove default `protocol` value as this lead to false errors in certain circumstances.
If `protocol` is not provided at host creation time, then raise an error and is not provided for an update then raise a warning about ignoring initiator update.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_host.py

##### ADDITIONAL INFORMATION
Provide example of how to use this fix

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
@@ -55,7 +55,6 @@ options:
     description:
     - Defines the host connection protocol for volumes.
     type: str
-    default: iscsi
     choices: [ fc, iscsi, nvme, mixed ]
   wwns:
     type: list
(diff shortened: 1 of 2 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
```diff
@@ -313,23 +320,26 @@ def _is_cbs(array, is_cbs=False):
...
             try:
                 array.set_host(module.params["name"], iqnlist=module.params["iqn"])
             except Exception:
                 array.delete_host(module.params["name"])
                 module.fail_json(msg="Setting of iSCSI IQN failed.")
-    if module.params["protocol"] in ["fc", "mixed"]:
+    if module.params["protocol"].lower() in ["fc", "mixed"]:
         if module.params["wwns"]:
             try:
+                module.warn("Here")
```
rdeenik, line 342: Remove this, was probably used for debugging.

rdeenik (dismissed):
lgtm, small nit

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_host.py
sdodsley: Done. Thanks for catching that.

PR #213, opened 2021-07-20
Add KMIP support

sdodsley:
##### SUMMARY
Add support KMIP server object management

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_kmip.py

##### ADDITIONAL INFORMATION
Testing notes: Create your own SSL cert and use this to test KMIP object management. 

**DO NOT MERGE UNTIL 1.10.0 IS READY TO GO**


rdeenik (approved):
LGTM. Maybe add a comment that the certificate needs to exist on the array and can be created using the `purecert create --certificate --key` command?

sdodsley:
It does fail with an error message stating that on KMIP creation, but admittedly not on update. I'll add that.
Why would I not tell them to use the `purefa_ceets` module??

rdeenik (approved):
Shipit

PR #224, opened 2021-09-14
Add NMVe NGUID to volume dict

sdodsley:
##### SUMMARY
Add the NVMe NGUID to volume dicts.
This will be added even if NVMe protocol is not being used by the volume.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_info.py


rdeenik (approved):
lgtm

PR #238, opened 2021-11-01
Add directory snapshot rename support

sdodsley:
##### SUMMARY
Add support for directory snapshot rename.
Requires Purity//FA 6.2.1 or higher and FA-Files enabled.
Add new parameters:
  `rename`: boolean - default False
  `new_client`: string
  `new_suffix`: string - must conform to suffix naming rules
Remove default value for `keep_for` to allow for rename and retention time changes together.
Add extra check for `keep_for` in the allowed range during snapshot creation.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_dirsnap.py

#### ADDITIONAL INFORMATION
`py-pure-client` minimum requirement changed to `1.19` as a minimum


rdeenik (approved):
shipit

PR #243, opened 2021-11-03
Add new global admins module and deprecate SSO module

sdodsley:
##### SUMMARY
Deprecate `purefa_sso` in favor of the new module `purefa_admin` which includes SSO along with new parameters

##### ISSUE TYPE
- Docs Pull Request
- New Module Pull Request

##### COMPONENT NAME
purefa_admin.py


rdeenik (approved):
Shipit (small not, ideally in the future we'd also be able to remove the settings using this module).

Issue #267, opened 2022-02-17
Quota Policies are not idempotent 


MilesTails:
**Describe the bug**
Unable to run ansible job multiple times with quota policy rules. On subsequent runs the job appears to try and create new rules rather than maintain idempotency.

**To Reproduce**
Data Model:
```yaml
fa_quota_policies:
  - name: "miles-quota"
    rules:
      - quota_enforced: "no"
        quota_limit: "1G"
        quota_notifications: "user"
        ignore_usage: "no"
      - quota_enforced: "yes"
        quota_limit: "2G"
        quota_notifications: "user"
        ignore_usage: "no"
```

Ansible task:
```yaml
- name: Create directory quota policies
  purestorage.flasharray.purefa_policy:
    name: "{{item.0.name}}"
    policy: quota
    quota_enforced: "{{item.1.quota_enforced | default('yes')}}"
    quota_limit: "{{item.1.quota_limit | mandatory }}"
    quota_notifications: "{{item.1.quota_notifications | default(omit)}}"
    ignore_usage: "{{item.1.ignore_usage | default('no') }}"
    fa_url: "{{ fa_url }}"
    api_token: "{{ fa_token }}"
  loop: "{{ fa_quota_policies | subelements('rules') }}"    
```

On second run receive error due to only one rule being allowed to enforce the quota.
```yaml
TASK [pure_fa_filesystem : Create directory quota policies] ********************
changed: [localhost] => (item=[{'name': 'miles-quota', 'rules': [{'quota_enforced': 'no', 'quota_limit': '1G', 'quota_notifications': 'user', 'ignore_usage': 'no'}, {'quota_enforced': 'yes', 'quota_limit': '2G', 'quota_notifications': 'user', 'ignore_usage': 'no'}]}, {'quota_enforced': 'no', 'quota_limit': '1G', 'quota_notifications': 'user', 'ignore_usage': 'no'}]) => {"ansible_loop_var": "item", "changed": true, "item": [{"name": "miles-quota", "rules": [{"ignore_usage": "no", "quota_enforced": "no", "quota_limit": "1G", "quota_notifications": "user"}, {"ignore_usage": "no", "quota_enforced": "yes", "quota_limit": "2G", "quota_notifications": "user"}]}, {"ignore_usage": "no", "quota_enforced": "no", "quota_limit": "1G", "quota_notifications": "user"}]}
[...]

Issue #266, opened 2022-02-17
Unable to create snapshot policy with multiple rules


MilesTails:
**Describe the bug**
Unable to create snapshot policy with multiple rules. Error claims that interval of the same policy must be in multiples . Confirmed that we are able to create the rule via the gui. I have reviewed the code and can't see this specific error message.

Also just on a side note to this, how do I map snapshot policies to directories? I had a look through the code and docs but I couldn't work out or see anywhere how to make that mapping?

**To Reproduce**
Data Model:
```yaml
fa_snapshot_policies:
  - name: "test-snapshot-pol"
    rules:
      - snap_client_name: "foo"
        snap_suffix: "miles"
        snap_every: 1440
        snap_keep_for: 1440
      - snap_client_name: "foo2"
        snap_suffix: "miles2"
        snap_every: 900
        snap_keep_for: 900
```

Ansible task:
```yaml
- name: Create snapshot policies
  purestorage.flasharray.purefa_policy:
    name: "{{item.0.name}}"
    policy: snapshot
    snap_client_name: "{{item.1.snap_client_name | mandatory}}"
    snap_suffix: "{{item.1.snap_suffix | default(omit)}}"
    snap_every: "{{item.1.snap_every | mandatory }}"
    snap_keep_for: "{{item.1.snap_keep_for | mandatory}}"
    snap_at: "{{item.1.snap_at | default(omit)}}"
    fa_url: "{{ fa_url }}"
    api_token: "{{ fa_token }}"
  loop: "{{ fa_snapshot_policies | subelements('rules') }}"    
```

On run receive error: 
`Failed to create new rule for Snapshot policy test-snapshot-pol. Error: Snapshot interval of the rules in the same policy must be multiples`

**Expected behavior**
It creates the second rule.

**Environment**
Ansible 2.10.2
Collection v1.12.1+policy monkey patch
Purity//FA - 6.1.10

sdodsley:
@MilesTails regarding your question on assigning directories to a snapshot - that piece is missing.
We will get it added when we fix this bug

sdodsley:
@MilesTails the error you are seeing is sort of correct.... It is coming Purity itself.
If I do this manually in the GUI I actually get 2 error messages:
`test_dir: Snapshot interval of the rules in the same policy must be multiples.
test_dir: "suffix" can only be set for at most one rule in the same policy.`
But we only report the first error.
[...]

PR #268, opened 2022-02-17
Fix quota policies idempotency and add snapshot policy directories

sdodsley:
##### SUMMARY
Bug from #267 where quotas were not idempotent causing additional quotas to be created and enforced quotas to fail even when already existing.

From #266 add missing feature to allow directories to be managed in a snapshot policy.
New example provided for this functionality

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_policy.py


rdeenik (approved):
shipit

PR #273, opened 2022-03-24
Purepg fixes

masteph:
##### SUMMARY
These commits fix some bugs related to usage of protection groups inside of pods.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pg
purefa_pgsnap

##### ADDITIONAL INFORMATION
- purefa_pg
-- renaming of protection groups forced the new name to be lower case
-- renaming of protection groups inside of pods did actually not do anything due to wrong indentation
- purefa_pgsnap
-- restore of snapped volumes to stretched pods was forbidden by code but is allowed in Pure UI since at least 6.1.13, probably even earlier

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
@@ -730,6 +730,7 @@ def main():
     array = get_system(module)
     pattern = re.compile("^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$")
     if module.params["rename"]:
+        module.params["rename"] = module.params["rename"].lower()
         if not pattern.match(module.params["rename"]):
             module.fail_json(
                 msg="Rename value {0} does not conform to naming convention".format(
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
@@ -730,7 +730,6 @@ def main():
     array = get_system(module)
     pattern = re.compile("^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$")
     if module.params["rename"]:
-        module.params["rename"] = module.params["rename"].lower()
         if not pattern.match(module.params["rename"]):
             module.fail_json(
                 msg="Rename value {0} does not conform to naming convention".format(
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
@@ -615,13 +615,13 @@ def update_pgroup(module, array):
                     rename = container + ":" + module.params["rename"]
             else:
                 rename = module.params["rename"]
-                renamed = True
-                if not module.check_mode:
-                    try:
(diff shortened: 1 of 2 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
```diff
@@ -730,7 +730,6 @@ def main():
     array = get_system(module)
     pattern = re.compile("^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$")
     if module.params["rename"]:
-        module.params["rename"] = module.params["rename"].lower()
```
sdodsley, line 733: We can't just ignore the case as Purity is case insensitive.
Given Ansible still supports Python2.7 we have to force lowercase for consistency across all the modules.
When Python 2.7 is deprecated by Ansible there are a number of modules that will have `lower()` replaced with `casefold()` to allow case insensitivity to be fully supported by the Pure modules

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pgsnap.py
```diff
@@ -307,15 +307,6 @@ def restore_pgsnapvolume(module, array):
         + "."
         + module.params["restore"]
     )
-    if "::" in module.params["target"]:
```
sdodsley, line 310: I need to check on which version this requirement changed and then we need to add in a API version check so that people running older versions of Purity still get this error.

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
```diff
@@ -615,13 +615,13 @@ def update_pgroup(module, array):
                     rename = container + ":" + module.params["rename"]
             else:
                 rename = module.params["rename"]
-                renamed = True
```
sdodsley, line 618: Good catch

sdodsley (changes requested):
I can only approve indent issue section of this PR.

See comments.

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
masteph: I see your point. On the other hand the UI allows creation and renaming in mixed case.
Maybe the introduction of a "lowercase_names" (bool, default True) option could be a solution to this?

sdodsley: There are many modules that would need this if we are to be consistent. In the interim we will need to stick to lowercase

sdodsley:
@masteph please check out this change to your PR. 
Hopefully this resolves your issues except the `lowercase` problem that will have to wait for Ansible to deprecate Python 2.7 support

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pg.py
sdodsley: Removing this change pending Python 2.7 deprecation

collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pgsnap.py
sdodsley: Added API version check

sdodsley:
All that remains of this is now the support for restoring stretched pods.

PR #275, opened 2022-04-07
fixed IPv6 netmask handling

kretuq:
Previously setting an IPv6 address resulted in a failure since FlashArray REST API only accepts prefix length for IPv6 addresses.
For legacy IP only decimal subnet notation is accepted, prefix length is not. 
Added a conditional to use prefix length if IPv6 is used and keep default subnet notation for legacy IP.
Since those parameters are passed to py-pure-client the issue could be fixed there aswell.
For now this workaround does it's job

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
+++ b/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
@@ -204,7 +204,7 @@ def update_interface(module, array, interface):
             ]:
                 module.fail_json(msg="Gateway and subnet are not compatible.")
         address = str(module.params["address"].split("/", 1)[0])
-        ip_version = str(IPAddress(address).version)
+    ip_version = str(IPAddress(address).version)
     if not module.params["mtu"]:
         mtu = interface["mtu"]
     else:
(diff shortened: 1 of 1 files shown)


collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_network.py
```diff
@@ -204,6 +204,7 @@ def update_interface(module, array, interface):
             ]:
                 module.fail_json(msg="Gateway and subnet are not compatible.")
         address = str(module.params["address"].split("/", 1)[0])
+        ip_version = str(IPAddress(address).version)
```
sdodsley, line 207: This needs to be moved in 4 spaces so that it is set even if the `address` parameter is not set, otherwise your change on line 231 will not have a value set for `ip_version`

sdodsley (changes requested):
Looks good, but a slight alteration is needed...

kretuq:
why did you close instead of merge?

sdodsley:
I thought I merged it.

Sorry. I'll fix that

Simon Dodsley
Director Telco & Open Source Integrations
Pure Storage Inc
Cell: (510) 516-2682

PR #292, opened 2022-05-20
Fix bind password idempotency

sdodsley:
##### SUMMARY
Add new parameter `force_bind_password` which defaults to `True`
If the bind user is not changed and `force_bind_password` is set `false` then this allows the
module to be idempotent.
`force_bind_password` is ignored if the bind user is changed and a password is required at this point

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_ds.py

### DIFF
--- a/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/292_fix_ds_password.yaml
+++ b/collections/ansible_collections/purestorage/flasharray/changelogs/fragments/292_fix_ds_password.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+ - purefa_ds - Add new parameter `force_bind_password` (default = True) to allow idempotency for module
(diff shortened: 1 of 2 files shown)


sanderdescamps:
I tested the improved module and it works perfectly. 
I'm only suggesting to add the `no_log=False` parameter to the argument spec of `force_bind_password`. Because otherwise Ansible returns an error telling the `no_log` parameter is not set. 

```
TASK [infra.infra_storage.pure_storage_install : Configure directory service on Pure Storage ] *************************************************
[WARNING]: Module did not set no_log for force_bind_password
ok: [pure1.domain.com -> localhost]
ok: [pure2.domain.com -> localhost]
```

sdodsley:
@sanderdescamps `no_log` added

PR #299, opened 2022-07-27
Fix pgsched issue for handlng of replication params

sdodsley:
##### SUMMARY
Fix issue when replication schedule would not correctly handle 0 for the `taregt_days` and `target_per_day` parameters

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsched.py


sirhc-chris:
handling of `days` follows the same suit (it assumes the default of `7` if either the parameter is unspecified or the value is set to `0`):
https://github.com/Pure-Storage-Ansible/FlashArray-Collection/blob/ce854b4b7a997b68eb20577214c100b51a57fead/collections/ansible_collections/purestorage/flasharray/plugins/modules/purefa_pgsched.py#L262-L268

this should be:
```
        if not module.params["days"]:
            if isinstance(module.params["days"], int):
                days = module.params["days"]
            else:
                days = current_snap["days"]
        else:
            if module.params["days"] > 4000:
                module.fail_json(msg="Maximum value for days is 4000")
            else:
                days = module.params["days"]
```

sdodsley:
@sirhc-chris updated patch to cover `days`. Please check

skeerthivasan (commented):
Validated the changes and ran tests its working fine.

PR #320, opened 2022-09-14
Ensure latest snapshot is complete before using for restore

sdodsley:
##### SUMMARY
Snap Restore can fail if the latest snapshot is still in the middle of processing and hasn't completed.
Ensure that when selecting `latest` we get the latest completed snapshot.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py


StephenEGibson:
This has resolved my issue. I just caught it in the middle of transitioning to the next cycle of the protection group schedule and it did not error out.

PR #436, opened 2023-08-14
Snap fix

### LINKED ISSUE #435: Remote snapshots are not deleted
**Describe the bug**
purefa_snap unable to destroy/eradicate remoe snapshots

**To Reproduce**
1. Create a remote snapshot
2. Destroy and eradicate remote snapshot from target array
3. Verify the snapshot still exists on target array

**Expected behavior**

Remote snapshot gets deleted

** Additional notes **
The problem is due to a possible issue with API 1.x, in which the volume search method returns an error if a remote snapshot is specified.

genegr:
##### SUMMARY

Fixed #435 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_snap

### DIFF
--- a/changelogs/fragments/436_snap_fix.yaml
+++ b/changelogs/fragments/436_snap_fix.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+ - purefa_snap - Fixed issue with remote snapshot retrieve. Mainly a workaround to an issue with Purity REST 1.x when remote snapshots are serached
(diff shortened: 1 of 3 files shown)


sdodsley (dismissed):
No position to test but I assume you have tested multiple scenarios...

sdodsley:
@genegr please add a release note

genegr:
Yep, tested all the combinations of automatically and user assigned postfix
and it worked. It is mainly a workaround to the response error returned by
the volumes API 1.x. A normal query for a remote snapshot returns an error
with the specific reason that is a remote volume snapshot, while with the
wildcard '*' it matches all the snapshots for the volumes, without taking
into consideration they are remote or local.

genegr:
Done. Added the proper changelog fragment file

sdodsley:
Changelog fragments needs to be a .yml fe

changelogs/fragments/436_snap_fix.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+ - purefa_snap - Fixed issue with remote snapshot retrieve. Mainly a workaround to an issue with Purity REST 1.x when remote snapshots are serached
```
sdodsley, line 2: Spelling

genegr: Amended

sdodsley (approved):

PR #459, opened 2023-09-28
fix errors collecting eradication config on new purity versions

### LINKED ISSUE #458: Using purefa_info fails on pure arrays on 6.4.10+ when there are unique values for the two eradication timers
**Describe the bug**
When using `purefa_info` on an array that is running Purity 6.4.10+ the new eradication timers (enabled and disabled) are returned as a single timer if they are the same. If they are different values a parameter error is returned.

tylergmuir:
##### SUMMARY

Fixes errors in collecting information from arrays that have different enabled and disabled eradication timers after upgrading to purity 6.4.10+ by returning the appropriate values.

Fixes #458 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

purefa_info.py

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/purefa_info.py
```diff
@@ -131,6 +131,7 @@
 POD_QUOTA_VERSION = "2.23"
 AUTODIR_API_VERSION = "2.24"
 SUBS_API_VERSION = "2.26"
+OBJECT_SAFE_MODE_API_VERSION = "2.26"
```
sdodsley, line 134: You can use `SUBS_API_VERSION` as this already define 2.26

plugins/modules/purefa_info.py
```diff
@@ -166,9 +167,17 @@ def generate_default_dict(module, array):
                 default_info["encryption_algorithm"] = encryption.data_at_rest.algorithm
                 default_info["encryption_module_version"] = encryption.module_version
             eradication = array_data.eradication_config
-            default_info["eradication_days_timer"] = int(
-                eradication.eradication_delay / SEC_TO_DAY
-            )
+            if OBJECT_SAFE_MODE_API_VERSION in api_version:
```
sdodsley, line 170: See previous comment

plugins/modules/purefa_info.py
```diff
@@ -166,9 +167,17 @@ def generate_default_dict(module, array):
...
+                default_info["eradication_safe_mode_disabled_days_timer"] = int(
+                    eradication.disabled_delay / SEC_TO_DAY
+                )
+                default_info["eradication_safe_mode_enabled_days_timer"] = int(
+                    eradication.enabled_delay / SEC_TO_DAY
+                )
+            else:
+                default_info["eradication_days_timer"] = int(
```
sdodsley, line 178: `eradication_delay` is still a valid option for REST 2.26, depending on how the array is configured, so I would not have this as an `else` and use `getattr` to provide a value of `None` if it isn't provided, then do the `SEC_TO_DAY` conversion if it does exist.

sdodsley (changes requested):
Thank you for your contribution.
A few changes required.
Also, please sign the CLA so we can accept your patch.

plugins/modules/purefa_info.py
tylergmuir: I did see that there was an existing variable for the same API version, but the reason I had opted to have a separate line was for 2 reasons. First, it was more descriptive as to what the check was for. There was existing precedence to this as there are multiple variables for the `1.14` and `1.16` versions.

The other reason was that the `purefa_eradication` module most likely will also need to be updated to support the additional fields. I looked at it briefly and while it isn't broken like the `purefa_info` module is in this case, it does limit the options available. In the case the the `purefa_eradication` module is also updated, it would be convenient to have the same variable name across the modules for the same purpose where the name `SUBS_API_VERSION` wouldn't mean anything useful in the `purefa_eradication` module context.

All of that said, I am open to changing it to just use the `SUBS_API_VERSION`, I just figured it was worth the discussion.

plugins/modules/purefa_info.py
tylergmuir: That is a good point. I will get this updated shortly.

plugins/modules/purefa_info.py
```diff
@@ -166,9 +166,18 @@ def generate_default_dict(module, array):
...
+                default_info["eradication_safe_mode_disabled_days_timer"] = int(
+                    eradication.disabled_delay / SEC_TO_DAY
+                )
+                default_info["eradication_safe_mode_enabled_days_timer"] = int(
+                    eradication.enabled_delay / SEC_TO_DAY
+                )
+            eradication_delay = getattr(eradication, "eradication_delay", None)
+            if eradication_delay is not None:
+                default_info["eradication_days_timer"] = int(
+                    eradication_delay / SEC_TO_DAY
```
sdodsley, line 179: this still needs `eradication.`

sdodsley (changes requested):

plugins/modules/purefa_info.py
sdodsley: this still needs addressing

sdodsley (changes requested):

plugins/modules/purefa_info.py
tylergmuir: Are you referring to not using `getattr` and instead doing this?
```
if eradication.eradication_delay is not None:
  default_info["erdaication_days_timer"] = int (
    eradication.eradication_delay / SEC_TO_DAY
  )
```

Or using `getattr` but inline in the if statement like this?
```
if getattr(eradication, "eradication_delay", None) is not None:
  default_info["erdaication_days_timer"] = int (
    eradication.eradication_delay / SEC_TO_DAY
  )
```

My implementation was just "extracting" the `eradication_delay` attribute to a variable to use.

sdodsley: My bad - misread the code. Looks good now

sdodsley (approved):

PR #465, opened 2023-09-30
Fixes error using force_bind_password

### LINKED ISSUE #464: Failure creating new ds configuration when not using force_bind_password
**Describe the bug**
If you have `force_bind_password: false` and there is no current directory services configuration, you will receive an error stating "Both bind user and bind password must be configured." even when providing both `bind_user` and `bind_password`.

**To Reproduce**
1. Have no current directory services configuration.
2. Attempt to configure a new directory services configuration with using `force_bind_password: false`.

**Expected behavior**
The directory services configuration should be applied without error.

tylergmuir:
##### SUMMARY

This fixes a logic error. The line before the if statement, the variables `bind_user` is being set to `module.params["bind_user"]` and then the if statement is checking is `bind_user` is equal to `module.params["bind_user"]`. This is causing the following code in the if statement to never run. Due to this, if you use the parameter `force_bind_password: false` when calling `purefa_ds` and there is no existing directories services configuration you receive an error.

Fixes #464 
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

purefa_ds

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/purefa_ds.py
+++ b/plugins/modules/purefa_ds.py
@@ -434,8 +434,8 @@ def update_ds_v6(module, array):
         base_dn = module.params["base_dn"]
         ds_change = True
     if module.params["bind_user"] != "":
-        bind_user = module.params["bind_user"]
-        if module.params["bind_user"] != current_ds.bind_user:
+        if module.params["bind_user"] != bind_user:
+            bind_user = module.params["bind_user"]
             password_required = True
             ds_change = True
         elif module.params["force_bind_password"]:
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_ds.py
```diff
@@ -435,7 +435,7 @@ def update_ds_v6(module, array):
         ds_change = True
     if module.params["bind_user"] != "":
         bind_user = module.params["bind_user"]
```
sdodsley, line 437: It looks to me that the actual fix required is to just move line 437 into the `if` block starting on 438. No need for the change to the `if` statement 

Please try that.

sdodsley (changes requested):

plugins/modules/purefa_ds.py
tylergmuir: I agree. My original code needlessly sets the variable to what it already is in the case that the `uri` parameter was set to the same thing as what the array was set to.

sdodsley (approved):

PR #470, opened 2023-10-02
fix missing bind password

### LINKED ISSUE #469: purefs_ds fails to apply the configuration if there is no bind_password set but there is a bind_user set
**Describe the bug**
In the GUI for the pure array, it allows you to set a bind user without setting the bind password. This configuration doesn't work, but it is allowed.

If you have an array in this state and you are using `force_bind_password: false` it doesn't detect that there is no password set so it doesn't attempt to include the `bind_password`.

**To Reproduce**
1. Have an array with the directory services configured with a bind username and no bind password and have the directory services disabled.
2. Attempt to configure the directory services with ansible to enable it and also have `force_bind_password: false`

**Expected behavior**
The expected behavior is that ansible would see that there is no password set and update the password.

tylergmuir:
##### SUMMARY

In the case that the directory services config is currently enabled, it will always have a `bind_password` set so there will be no change in expected action. But in the case that the directory services is currently disabled and Ansible it attempting to enable it and the `bind_password` isn't currently set, the `bind_password` is required to be able to enable the directory services.

This is all ignored in the case that `force_bind_password` is enabled as it will already set the flag to update the password as a part of the change.

Fixes #469 
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

purefa_ds
##### ADDITIONAL INFORMATION

```paste below

```


sdodsley (approved):
Looks good.
Thank you

PR #474, opened 2023-10-03
Multiple network module fixes/changes

### LINKED ISSUE #391: Requesting purefa_network support for bond / lacp / virtual interfaces
Equivalent configuration option to match the CLI commands:
`purenetwork eth create lacpbond [--subinterfacelist SUBINTERFACELIST] INTERFACE`
`purenetwork eth create vif [--address ADDRESS] { --subnet SUBNET | --subinterfacelist SUBINTERFACELIST } INTERFACE...`

**Additional context**
There are currently no options under the **purefa_network** module to fully configure these type of interfaces. There is a way to specify a vif interface but only with `address` attribute and not `subnet` nor `subinterfacelist`.

### LINKED ISSUE #466: some issues regarding purefa_network and purefa_subnet
My customer is testing ansible environment for production use, and found some issues.
Could you please fix these?

- purefa_network
01.configure IPv6 address or gateway
02.change IPv6 address or gateway
03.delete IPv6 address or gateway

- purefa_subnet
04.remove **IPv4** gateway
05.create IPv6 subnet
06.change IPv6 prefix or gateway

•             Ansible version: ansible [core 2.15.4]
•             FlashArray Collection version: 1.21.0
•             Purity version: 6.3.13
•             FlashArray module causing the failure: purefa_network and purefa_subnet
•             Name of the interface and the service(s) enabled on the interface: ct0.eth5
•             An example task that fails
•             Response from ansible-playbook -vvvv when running a failing task: see attached file

[01.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780683/01.log)
[02.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780684/02.log)
[03.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780685/03.log)
[04.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780686/04.log)
[05.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780687/05.log)
[06.log](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12780688/06.log)

### LINKED ISSUE #472: unable to update network interface with no IP address set
**Describe the bug**
When you have an interface that you want to update without an IP address set, say in the case you are changing the services. The purefa_network module fails with an error that state "failed to detect a valid IP". A workaround is to set the address to `0.0.0.0/0` but this results in the module reporting `changed` every time.

**To Reproduce**
1. Have an interface that doesn't have an IP address set. 
2. Attempt to set only the services and/or MTU of the interface and do not specify the `address` or `gateway` parameters.

**Expected behavior**
The expected behavior is that the network interface would be updated as it is a valid state to have no address assigned.

### LINKED ISSUE #473: network port reports as changed when removing the address or gateway
**Describe the bug**
When using `purefa_network`, if you have `address` set to `0.0.0.0/0` or `gateway` set to 0.0.0.0 to remove either of those configurations from the network on the array, it always reports back `changed` regardless of if anything was changed.

**To Reproduce**
1. Have a network port without an IP address or gateway configured.
2. Use the `purefa_network` module with `state: present`, `address: 0.0.0.0/0`, and `gateway: 0.0.0.0`.

**Expected behavior**
The expected behavior is that if nothing on the port is changed, the status should report `ok` instead of `changed`.

sdodsley:
##### SUMMARY
Add support for `interface` parameter with options of `vif, lacp`. New, associated, parameters of `subinterfaces` for interface type of `vif` and `subordinates` of interface type `lacp`.

Add parameters `enabled` (default True) and `subnet`.

Closes #466
Closes #473 
Closes #472 
Closes #391 

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_network.py

### DIFF
--- a/plugins/modules/purefa_network.py
+++ b/plugins/modules/purefa_network.py
@@ -67,6 +67,39 @@ options:
     type: list
     choices: [ "replication", "management", "ds", "file", "iscsi", "scsi-fc", "nvme-fc", "nvme-tcp", "nvme-roce", "system"]
     version_added: '1.15.0'
+  interface:
+    description:
+      - Type of interface to create if subinterfaces is supplied
+    type: str
+    choices: [ "vif", "lacp" ]
+    version_added: '1.22.0'
+  subordinates:
+    description:
+     - List of one or more child devices to be added to a LACP interface
+     - Subordinates must be on the same controller, therefore the full device needs
+       to be provided.
+    type: list
+    elements: str
+    version_added: '1.22.0'
+  subinterfaces:
+    description:
+     - List of one or more child devices to be added to a VIF interface
+     - Only the 'eth' name needs to be provided, such as 'eth6'. This interface on
+       all controllers will be assigned to the interface.
+    type: list
+    elements: str
+    version_added: '1.22.0'
+  subnet:
+    description:
+     - Name of the subnet which interface is to be attached
+    type: str
+    version_added: '1.22.0'
+  enabled:
+    description:
+    - State of the network interface
+    type: bool
+    default: true
+    version_added: '1.22.0'
 extends_documentation_fragment:
     - purestorage.flasharray.purestorage.fa
 """
@@ -117,14 +150,21 @@ RETURN = """
 """
 
 try:
-    from netaddr import IPAddress, IPNetwork
+    from netaddr import IPAddress, IPNetwork, valid_ipv4, valid_ipv6
 
     HAS_NETADDR = True
 except ImportError:
     HAS_NETADDR = False
 
 try:
-    from pypureclient.flasharray import NetworkInterfacePatch
+    from pypureclient.flasharray import (
+        NetworkInterfacePatch,
+        NetworkInterfacePost,
+        NetworkinterfacepostEth,
+        NetworkinterfacepatchEth,
+        FixedReferenceNoId,
+        ReferenceNoId,
+    )
 
     HAS_PYPURECLIENT = True
 except ImportError:
@@ -154,8 +194,7 @@ def _get_fc_interface(module, array):
     if interface_list.status_code == 200:
         interface = list(interface_list.items)[0]
         return interface
-    else:
-        return None
+    return None
 
 
 def _get_interface(module, array):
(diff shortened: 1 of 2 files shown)


tylergmuir:
From my testing, this doesn't appear to fix #473. I think the issue is that the `address` that is being past in is `0.0.0.0/0` but the pure array returns `None`. Same goes for the gateway.

sdodsley:
I'll take another look at it

sdodsley:
@tylergmuir I think that gets it now. Please let me know if your testing is successful or not.

tylergmuir:
@sdodsley That did fix the reporting on the interfaces. But I just tested updating an existing `vif` and it failed. Collecting more info to provide now and will have that to you shortly.

plugins/modules/purefa_network.py
```diff
@@ -268,53 +378,117 @@ def update_interface(module, array, interface):
...
             )
         if not module.check_mode:
             try:
+                array.set_network_interface(
+                    interface["name"], enabled=new_state["enabled"]
+                )
                 if new_state["gateway"] is not None:
                     array.set_network_interface(
                         interface["name"],
+                        enabled=new_state["enabled"],
```
tylergmuir, line 463: This causes the `try` block to file. I think you had discovered this which is why the `set_network_interface` call is made just above to enable the interface.

I think to fix this, you just need to remove the `enabled=new_state["enabled"],` line.

plugins/modules/purefa_network.py
```diff
@@ -227,35 +266,106 @@ def update_fc_interface(module, array, interface, api_version):
...
         "services": sorted(interface["services"]),
+        "slaves": sorted(interface["slaves"]),
     }
+    array6 = get_array(module)
+    subinterfaces = current_state["slaves"]
+    if module.params["subinterfaces"]:
+        if sorted(module.params["subinterfaces"]) != current_state["slaves"]:
+            subinterfaces = _create_subinterfaces(module, array6)
+        else:
+            subinterfaces = current_state["slaves"]
```
tylergmuir, line 332: There are a couple of issues here. Assuming you are following the description of the `subinterfaces` parameter and only providing the "eth" number, the if statement will always equate to true. This is because `module.params["subinterfaces"]` returns just the eth numbers and `current_state["slaves"]` returns the full interface name including the controller. 

The second issue is that the `current_state["slaves"]` is sorted. but the response from `_create_subinterfaces` is not sorted. So later on when evaluating if the `current_state` and the `new_state` are the same, it always fails if you are using more than 1 interface on each controller. This could be fixed here when setting `subinterfaces` or in the return from the `_create_subinterfaces` function.

plugins/modules/purefa_network.py
```diff
@@ -227,35 +266,106 @@ def update_fc_interface(module, array, interface, api_version):
...
+            "0.0.0.0",
+            "::",
+        ]:
+            if module.params["gateway"] not in IPNetwork(module.params["address"]):
                 module.fail_json(msg="Gateway and subnet are not compatible.")
             elif not module.params["gateway"] and interface["gateway"] not in [
                 None,
                 IPNetwork(module.params["address"]),
             ]:
                 module.fail_json(msg="Gateway and subnet are not compatible.")
```
tylergmuir, line 365: This elif statement is never equating to true. The preceding if statement on line 355, one of the checks is if `module.params["gateway"]` exists so the latter check on line 361 will always be false. 

I believe the highlighted section should be un-indented once so the elif statement is a part of the if statement on line 355 and not within that if statement.

tylergmuir (commented):
I believe I found the reason for the error I was getting when updating an existing `vif` interface. Plus I found an additional couple of logic errors that would cause the reporting to be incorrect in some cases.

plugins/modules/purefa_network.py
sdodsley: Fixed

plugins/modules/purefa_network.py
sdodsley: True - Fixed

plugins/modules/purefa_network.py
sdodsley: Fixed

sdodsley:
@tylergmuir @ryuyamag I think this now fixes all your concerns.

tylergmuir:
@sdodsley It looks like the configuration for the `vif` is applying correctly now, but it is always reporting as changed.

sdodsley:
@tylergmuir can you give an example task that is reporting idempotency incorrectly

tylergmuir:
@sdodsley this is the basis of what I am doing 
```
- name: Configure vif for file
  purestorage.flasharray.purefa_network:
    name: filevif
    servicelist:
      - file
    interface: vif
    subinterfaces:
      - eth4
      - eth5
    address: "{{ ip_cidr }}"
    gateway: "{{ gateway }}"
    mtu: 1500
```

sdodsley:
@tylergmuir ok - fixed it... Hopefully this is the last iteration. I never want to see this module again :)

tylergmuir:
@sdodsley It works! I ran a few more tests with different configs and it did what I expected it to do each time.

ryuyamag:
@sdodsley

very sorry, it seems that two functions do not work.

 1. clear address ipv4&ipv6
  failed to detect a valid IP

 2. clear gateway ipv4&ipv6
  changed:true appears, but it's not changed.

I have attached the log file
[network2.zip](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12811548/network2.zip)

sdodsley:
@ryuyamag I think I got it now. Please try all your tests and let me know how it goes.
@tylergmuir please confirm this last fix didn't break anything in your tests.

ryuyamag:
@sdodsley 
I had to modify some Playbooks this time, but I can confirm that everything works fine.
thank you.

tylergmuir:
@sdodsley looks good on my side too.

PR #483, opened 2023-10-16
Fix pgsnap issue not replicating correctly

### LINKED ISSUE #478: Replication issue with purefa_pgsnap
If I set "now:true" and "remote:false" in purefa_pgsnap, replication will not be performed for the created snapshot.
If both are true, it works fine.

The description of "remote" is below, if this param is "false", I think that replication task will queued and replicate later.
> Force immeadiate snapshot to remote targets
> Choices:
>   false ← (default)
>   true

• Ansible version: ansible [core 2.15.4]
• FlashArray Collection version: 1.21.0
• Purity version: 6.3.13
• FlashArray module causing the failure: purefa_pgsnap
• Response from ansible-playbook -vvvv when running a failing task: see attached file
[20231012.zip](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/12881864/20231012.zip)

sdodsley:
##### SUMMARY
Fixes #478 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py

### DIFF
--- a/changelogs/fragments/483_missing_replicate.yaml
+++ b/changelogs/fragments/483_missing_replicate.yaml
@@ -1,0 +1,2 @@
+breaking_changes:
+ - purefa_pgsnap - `now` and `remote` are now mutually exclusive.
--- a/plugins/modules/purefa_pgsnap.py
+++ b/plugins/modules/purefa_pgsnap.py
@@ -224,7 +224,7 @@ def get_pgroupvolume(module, array):
 
 
 def get_rpgsnapshot(module, array):
-    """Return iReplicated Snapshot or None"""
+    """Return Replicated Snapshot or None"""
     try:
         snapname = (
             module.params["name"]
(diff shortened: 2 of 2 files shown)


ryuyamag:
Considering the modified code, is it correct to have "replicate=true" in case of "now:false remote:true"?

This patch did not work in my environment. The fix may only be included in arrayv6.

sdodsley:
@ryuyamag I think the issue is that `remote` and `now` should be mutually exclusive and you only need to set either to True, with False being the default. Thereofre there are only 2 options possible.

sdodsley:
@ryuyamag please try this version.

ryuyamag:
I tested with this patch, I think this issue was fixed and can be closed.

However, "replicate" and "replicate_now" seemed to have the same behavior on RestAPI side.
"replicate" will not be queued up and the replicating at the same time as earlier replication sessions.

--
pureeng@japan-pslab-ct0:~$ purepgroup list --snap --transfer array1:yamapg
Name                Source         Created                  Started                  Completed  Progress  Data Transferred  Physical Bytes Written
array1:yamapg.p2    array1:yamapg  2023-10-19 11:25:22 JST  2023-10-19 11:25:22 JST  -          7.50%     553.80M           406.01M
array1:yamapg.p1    array1:yamapg  2023-10-19 11:23:54 JST  2023-10-19 11:23:54 JST  -          7.54%     316.98M           245.07M
array1:yamapg.1174  array1:yamapg  2023-10-19 11:22:27 JST  2023-10-19 11:22:27 JST  -          7.68%     2.75G             2.59G
--

sdodsley:
@ryuyamag all we can do with Ansible is use the API endpoints. If they are not doing what they should, then that would be an issue for Purity engineering to look into.

PR #488, opened 2023-10-20
Enabled to eradicate destroyed pgsnap

momokahirai:
##### SUMMARY
Enabled to eradicate destroyed protection group snapshots.
Added "not pgsnap_deleted" to the conditional expression so that "eradicate_pgsnapshot" would be executed for already-deleted snapshots. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py


sdodsley (approved):
Thank you for this

PR #498, opened 2023-11-02
Allow new PG to be created with both target and objects specified

### LINKED ISSUE #494: volume is not added with purefa_pg
When creating a PG with purefa_pg, the result seems fine.
But when I look at the actual array's PG, the target is set correctly, but the volume is not added to the members of.

```
changed: [localhost] => {
    "changed": true,
    "invocation": {
        "module_args": {
            "api_token": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "enabled": true,
            "eradicate": false,
            "fa_url": "10.229.192.60",
            "host": null,
            "hostgroup": null,
            "name": "testpg1",
            "rename": null,
            "safe_mode": false,
            "state": "present",
            "target": [
                "japan-pslab"
            ],
            "volume": [
                "testvol1"
            ]
        }
    }
}

```

•	Ansible version : core 2.15.5
•	FlashArray Collection version: 1.22.0
•	Purity version: 6.3.13
•	FlashArray module causing the failure: purefa_pg
•	An example task that fails: create new pg
•	Response from ansible-playbook -vvvv when running a failing task: please see attached
[purefa_pg_create.yml.zip](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/13226070/purefa_pg_create.yml.zip)

sdodsley:
##### SUMMARY
Previously when creating a new PG with `target` specified, the module ignored any other requests such as `volumes`, `hosts`, etc. 
This resolves that issue,
Closes #494 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pg.py

### DIFF
--- a/changelogs/fragments/498_fix_pg_creation.yaml
+++ b/changelogs/fragments/498_fix_pg_creation.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+ - purefa_pg - Allows a protection group to be correctly created when `target` is specified as well as other objects, such as `volumes` or `hosts`
--- a/plugins/modules/purefa_pg.py
+++ b/plugins/modules/purefa_pg.py
@@ -374,68 +374,65 @@ def make_pgroup(module, array):
                 module.fail_json(
                     msg="Creation of pgroup {0} failed.".format(module.params["name"])
                 )
+    if not module.check_mode:
+        try:
+            if module.params["target"]:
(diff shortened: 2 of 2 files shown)


ryuyamag:
Thank you, this patch resolved the issue.

PR #512, opened 2023-11-27
Fix clearing DS certificate

### LINKED ISSUE #514: unable to set ds to absent if it is currently absent
**Describe the bug**
When the directory services configuration has already be set to a state of absent, if you rerun the task it fails due to the `bind_password` not being set.

**To Reproduce**
1. Have a purefa_ds task where you set the state to absent.
2. Rerun the task on the same array.
3. See error

**Expected behavior**
The task should report as OK because the directory service config is already absent.

sdodsley:
##### SUMMARY
Fixes issue with clearing a management ceet with empty string
Fixes #514 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
purefa_ds.py

### DIFF
--- a/plugins/modules/purefa_ds.py
+++ b/plugins/modules/purefa_ds.py
@@ -443,7 +443,7 @@ def update_ds_v6(module, array):
     current_ds = dirserv
     if module.params["uri"] and current_ds.uris is None:
         password_required = True
-    if current_ds.uris != module.params["uri"]:
+    if module.params["uri"] and current_ds.uris != module.params["uri"]:
         uris = module.params["uri"]
         ds_change = True
     else:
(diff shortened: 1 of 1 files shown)


tylergmuir:
@sdodsley I opened an issue on purefa_ds (#514). I didn't open a PR since you already had one to fix a bug in the same module. I had a couple of thoughts on a fix. First was changing this:
```
        if state == "absent" and dirserv.uris != []:
            delete_ds_v6(module, arrayv6)
```

to this:

```
        if state == "absent":
            if dirserv.uris != []:
                delete_ds_v6(module, arrayv6)
            else:
                module.exit_json(changed=False)
```

at about line 620 should fix it.

But that would only fix the issue if we can assume that there is no DS configuration if there is no URI set. I didn't check to see if this is a good assumption or not or if you could set the rest of the configuration without the URI, so we should check against all of the attributes.

If you want to leave this for a separate PR, that is also fine with me.

ryuyamag:
@sdodsley 
Appreciate the additional update.
Clearing the cert is now working.

But, "check_peer" cannot be change to true or false even if there is an existing certificate.
It was possible in previous version.
[purefa_ds_disablepeer.yml.zip](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/files/13482329/purefa_ds_disablepeer.yml.zip)

ryuyamag:
I tried this version but result is same as the 4bab093.

Clearing the cert is now working.
But, "check_peer" cannot be change to true or false even if there is an existing certificate.

sdodsley:
@ryuyamag I think I fixed it .

ryuyamag:
@sdodsley 
I confirmed it fixed. Thanks a lot.

PR #523, opened 2024-01-03
Fix NFS policies and add SMB UBE

sdodsley:
##### SUMMARY
Fix issue where Post called instead of Patch for NFS policies
Add SMB user based enumeration parameter
Remove default setting for `nfs_version`.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_policy.py

### DIFF
--- a/changelogs/fragments/523_nfs_policies.yaml
+++ b/changelogs/fragments/523_nfs_policies.yaml
@@ -1,0 +1,5 @@
+bugfixes:
+ - purefa_policy - Fix incorrect call of psot instead of patch for NFS policies
+minor_changes:
+ - purefa_policy - Add SMB user based enumeration parameter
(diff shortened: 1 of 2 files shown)


genegr (dismissed):
SMB: create policy, add rules to policy, delete rules, update policy, delete policy all worked as expected.
NFS policy creation fails with any combinations of initial parameters. The execution creates the policy but with default values and without an initial rule (if the rule is specified). A simple play like the one in the examples fails, leaving the policy created with dafault vaules:

```
- name: Create an empty NFS policy with no rules
  purestorage.flasharray.purefa_policy:
    name: export1
    policy: nfs
    fa_url: 10.10.10.2
    api_token: e31060a7-21fc-e277-6240-25983c6c4592
```

genegr (approved):
Tested and also NFS policies management works.
Bump the patch release.

PR #614, opened 2024-08-13
use purefa_argument_spec()

### LINKED ISSUE #613: In purefa_token, main doesn't pull in the module defautls
**Describe the bug**
The purefa_token module main() should grab purefa_argument_spec() and the update() the spec with module parameters, but it does not. This causes what looks like a cert error, but is actually a keyerror about disable_warnings. disable_warnings is defined in purefa_argument_spec globally.

**To Reproduce**
Use 1.29.0 or greater of the collection and try to run the purefa_token module.

**Expected behavior**
Expect the token operation to succeed.

**Additional context**
Will submit PR.

lotaris-xx:
##### SUMMARY
Fix purefa_token to call purefa_argument_spec()  to grab global module parameters.

Fixes #613 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_token


lotaris-xx:
I'm unable to sign the CLA from our network.

sdodsley:
I suggest you abandon this and use #615 instead

PR #636, opened 2024-09-11
Fixed issue of unable to move volume from pod to vgroup

skeerthivasan:
##### SUMMARY

Unable to move volume from pod to vgroup.
Fixed by adding required condition to avoid checking vgroup as pod.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

purefa_volume.py


sdodsley (changes requested):
In this module's context a vgroup is a volume group. You are concerned about a protection group, which is never checked in this module.
You will need to add in a call to check if the volume is in a protection group.

sdodsley (approved):
Ignore previous comment - confused PRs

PR #638, opened 2024-09-12
Fix issue with pod deletion

sdodsley:
##### SUMMARY
Fixed an issue where a pod cannot be deleted if it has contents even if `delete_contents` was specified.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pod.py

### DIFF
--- a/plugins/modules/purefa_pod.py
+++ b/plugins/modules/purefa_pod.py
@@ -625,7 +625,9 @@ def delete_pod(module, array):
     changed = True
     if not module.check_mode:
         res = array.patch_pods(
-            names=[module.params["name"]], pod=PodPatch(destroyed=True)
+            names=[module.params["name"]],
+            pod=PodPatch(destroyed=True),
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_pod.py
```diff
@@ -625,7 +625,9 @@ def delete_pod(module, array):
     changed = True
     if not module.check_mode:
         res = array.patch_pods(
-            names=[module.params["name"]], pod=PodPatch(destroyed=True)
+            names=[module.params["name"]],
+            pod=PodPatch(destroyed=True),
+            eradicate_contents=module.params["delete_contents"],
```
skeerthivasan, line 630: For pod deletion to work fine, should change the eradicate_contents parameter to destroy_contents.

sdodsley: Agreed

skeerthivasan (approved):

PR #641, opened 2024-09-16
Fix keep_for parameter

sdodsley:
##### SUMMARY
When no `keep_for` is specified, the value defaulted to 0, however this should have been `None`.
Also fix issues with recovery of snapshots and correct application of `keep_for`
Add `time_remaining` for non-deleted directory snapshots in info module.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_dirsnap.py
purefa_info.py


skeerthivasan (commented):
@sdodsley , For the fix to work properly, we should add this condition hasattr(snap_detail, '.time_remaining') to line 302 & 312

skeerthivasan (approved):

PR #642, opened 2024-09-17
Errors out when pgroup and add_to_pgs used incorrectly

skeerthivasan:
##### SUMMARY
Errors out when pgroup and add_to_pgs used incorrectly
Below purity version 6.3.4 should use pgroup and above 6.3.4 should use add_to_pgs.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_volume.py

### DIFF
--- a/plugins/modules/purefa_volume.py
+++ b/plugins/modules/purefa_volume.py
@@ -319,6 +319,12 @@ try:
 except ImportError:
     HAS_PURESTORAGE = False
 
+HAS_PACKAGING = True
+try:
+    from packaging import version
+except ImportError:
+    HAS_PACKAGING = False
+
 import re
 import time
 from ansible.module_utils.basic import AnsibleModule
@@ -334,7 +340,6 @@ from ansible_collections.purestorage.flasharray.plugins.module_utils.common impo
 from ansible_collections.purestorage.flasharray.plugins.module_utils.version import (
     LooseVersion,
 )
-from packaging import version
 
 QOS_API_VERSION = "1.14"
 VGROUPS_API_VERSION = "1.13"
--- a/plugins/modules/purefa_volume.py
+++ b/plugins/modules/purefa_volume.py
@@ -334,6 +334,7 @@ from ansible_collections.purestorage.flasharray.plugins.module_utils.common impo
 from ansible_collections.purestorage.flasharray.plugins.module_utils.version import (
     LooseVersion,
 )
+from packaging import version
 
 QOS_API_VERSION = "1.14"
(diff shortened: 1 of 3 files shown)


sdodsley (changes requested):
Version number are not used in the Ansible modules - only REST versions, so you will need to find the REST version that came in with 6.3.4 - see https://wiki.purestorage.com/display/psw/REST+API+Versions+vs+Purity+Versions
There is no need to import packaging as we have a function in the collect to do version comparisons already - its a clone of LooseVersion. Check out some of the other module to see how to importnad use it.

plugins/modules/purefa_volume.py
```diff
@@ -1048,11 +1050,26 @@ def copy_from_volume(module, array):
     )
 
 
+def pg_check(module, array):
+    purity_version = array.get()["version"]
```
sdodsley, line 1054: Use the REST version instead of the Purity version for consistency. The REST version you need for 6.3.4 is 2.16, already defined as `DEFAULT_API_VERSION`.
The error message can reference the Purity version as that is what users will recognize.
You get the current rest version with `api_version = array.get_rest_version()`

sdodsley (changes requested):

plugins/modules/purefa_volume.py
skeerthivasan: Modified according to suggestions.

plugins/modules/purefa_volume.py
```diff
@@ -1048,11 +1050,23 @@ def copy_from_volume(module, array):
     )
 
 
+def pg_check(module, array):
+    rest_version = array.get_rest_version()
+    if LooseVersion(rest_version) >= LooseVersion("2.16") and module.params["pgroup"]:
```
sdodsley, line 1055: use the pre-defined `DEFAULT_API_VERSION`

sdodsley (changes requested):

skeerthivasan:
> This won't work. You are using `array.get_rest_version()` which uses REST 2. The `array` being passed into the function is the REST 1 version. You only need to pass `module` as the only parameter to the function and then define `array` from the REST 2 SDK using `array = get_array(module)`. This is because this module is a hybrid and uses both REST 1 and REST2 calls. If you look you will see reference to array6 all over where I'm differentiating between the two when they are both used in the same function.

@sdodsley , Strange it didn't error out during test, let me re-run and come back on this. Thanks.

sdodsley (dismissed):

PR #676, opened 2024-12-04
Fixes moving protected volume bug for volume group

### LINKED ISSUE #674: purefa_volume does not allow to move volume from vgroup if it's part of pgroup
Starting from collection version 1.32.0 purefa_volume module does not allow to move volume from volume group if it's part of protection group.
It returns "Cannot move volume in protection groups" error. 
Why this limitation was added to collection when it is possible to do that in array GUI (Purity 6.8.5)?

skeerthivasan:
##### SUMMARY
Bugfix done for issue of moving protected volume into volume group

Fixes #674 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_volume.py

### DIFF
--- a/changelogs/fragments/676_volume_vg_fix.yaml
+++ b/changelogs/fragments/676_volume_vg_fix.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+ - purefa_volume - Fixes issue of moving protected volume into volume group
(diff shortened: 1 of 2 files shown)


sdodsley (approved):
Looks good. LGTM

PR #679, opened 2024-12-06
Add realms support

sdodsley:
##### SUMMARY
Add new module to support Realms
Requires Purity//FA 6.6.11 or higher

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_realm.py


plugins/modules/purefa_realm.py
```diff
@@ -0,0 +1,484 @@
...
+        bw_qos_size = int(human_to_bytes(module.params["bw_qos"]))
+        if int(human_to_real(module.params["iops_qos"])) in range(
+            100, 100000000
+        ) and bw_qos_size in range(1048576, 549755813888):
+            changed = True
+            if not module.check_mode:
+                res = array.patch_realms(
+                    names=[module.params["name"]],
+                    realm=RealmPatch(
+                        iops_limit=module.params["iops_qos"],
```
skeerthivasan, line 273: iops_limit & bandwidth_limit should be in dict format & it should be assigned to 'qos' key.
This is leading to attribute error, should be corrected in all occurences of iops_limit & bandwidth_limit.
Eg: 
"qos": {
"bandwidth_limit": 1048576,
"iops_limit": 100
}

plugins/modules/purefa_realm.py
```diff
@@ -0,0 +1,496 @@
...
+
+    module.exit_json(changed=changed)
+
+
+def recover_realm(module, array):
+    """Recover Realm"""
+    changed = True
+    if not module.check_mode:
+        res = array.patch_realms(
+            module.params["name"], realm=RealmPatch(destroyed=False)
```
skeerthivasan, line 398: kwargs 'names' is missing, it should be names=[module.params["name"]]

PR #680, opened 2024-12-06
Update pods to support realms and change min reqs

sdodsley:
##### SUMMARY
Due to Realms support, need to deprecate a couple of pod function and replace with newer supporting realms.
Update docs and minimum requirements for the Collection.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
purefa_pod.py
README.md
requirements.txt

### DIFF
--- a/plugins/modules/purefa_pod.py
+++ b/plugins/modules/purefa_pod.py
@@ -198,6 +198,7 @@ from ansible_collections.purestorage.flasharray.plugins.module_utils.version imp
 
 POD_QUOTA_VERSION = "2.23"
 THROTTLE_VERSION = "2.31"
+MEMBERS_VERSION = "2.36"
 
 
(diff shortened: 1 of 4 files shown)


plugins/modules/purefa_pod.py
```diff
@@ -607,7 +607,7 @@ def stretch_pod(module, array):
             if not module.check_mode:
                 try:
                     array.delete_pods_arrays(
```
skeerthivasan, line 609: delete_pods_arrays should be changed to delete_pods_members.
Need to handle backward compatibility for older version array.

bsushanthkumar (commented):
LGTM

skeerthivasan (approved):
Good to merge

PR #684, opened 2024-12-13
Deprecate VNC feature

sdodsley:
##### SUMMARY
From Purity//FA 6.8.0 VNC feature for apps is deprecated as //RUN platform is transitioned to containers.
VNC module will raise a warning and fail cleanly if necessary
Info module will replace the VNC enabled boolean response with `"Deprecated"` as necessary

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_vnc.py
purefa_info.py

### DIFF
--- a/plugins/modules/purefa_vnc.py
+++ b/plugins/modules/purefa_vnc.py
@@ -86,12 +86,16 @@ try:
 except ImportError:
     HAS_PURESTORAGE = False
 
-
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.purestorage.flasharray.plugins.module_utils.purefa import (
     get_array,
     purefa_argument_spec,
 )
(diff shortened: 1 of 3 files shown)


plugins/modules/purefa_vnc.py
```diff
@@ -157,7 +161,10 @@ def main():
         module.fail_json(msg="py-pure-client sdk is required for this module")
 
     array = get_array(module)
-
+    api_version = array.get_rest_version()
+    if LooseVersion(MAX_API_VERSION) <= LooseVersion(api_version):
+        module.warn(msg="VNC feature deprecated from Purity//FA 6.8.0 and higher")
```
skeerthivasan, line 166: module.warn doesn't need 'msg' variable, should drop it and pass the string directly.

bsushanthkumar:
LGTM

PR #697, opened 2025-01-20
Add Fusion support to pods

sdodsley:
##### SUMMARY
Add Fusion support with `context` parameter.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
purefa_pods.py

### DIFF
--- a/plugins/modules/purefa_pod.py
+++ b/plugins/modules/purefa_pod.py
@@ -116,6 +116,14 @@ options:
     type: bool
     default: False
     version_added: '1.29.0'
+  context:
+    description:
+    - Name of fleet member on which to perform the volume operation.
+    - This requires the array receiving the request is a member of a fleet
+      and the context name to be a member of the same fleet.
+    type: str
+    default: ""
+    version_added: '1.33.0'
 extends_documentation_fragment:
 - purestorage.flasharray.purestorage.fa
 """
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_pod.py
```diff
@@ -199,74 +207,98 @@
...
+CONTEXT_VERSION = "2.38"
 
 
 def get_pod(module, array):
     """Return Pod or None"""
-    try:
-        return list(
+    api_version = array.get_rest_version()
+    if LooseVersion(CONTEXT_VERSION) <= LooseVersion(api_version):
+        return bool(
```
skeerthivasan, line 217: failure cases also provides error response dict, since its not empty dict bool considering it as true which is causing trouble, should be handled differently

plugins/modules/purefa_pod.py
```diff
@@ -116,6 +116,13 @@
     type: bool
     default: False
     version_added: '1.29.0'
+  context:
+    description:
+    - Name of fleet member on which to perform the volume operation.
+    - This requires the array receiving the request is a member of a fleet
+      and the context name to be a member of the same fleet.
+    type: str
```
skeerthivasan, line 124: missing default="", please set

sdodsley:
@skeerthivasan can you ensure that the QA checks cover setting of pod failover preferences. I think we are good and #710 will be fixed with this PR, but need you to check.

skeerthivasan:
> @skeerthivasan can you ensure that the QA checks cover setting of pod failover preferences. I think we are good and #710 will be fixed with this PR, but need you to check.

@sdodsley , Yeah it has been verified, we are good

skeerthivasan (approved):
Good to merge

PR #708, opened 2025-01-29
Fusion and REST 2 for pgsched

sdodsley:
##### SUMMARY
Convert to REST v2 and add Fusion support with `context` parameter.

##### ISSUE TYPE
- Feature Pull Request
- REST 2 Pull Request

##### COMPONENT NAME
purefa_pgsched.py

### DIFF
--- a/plugins/modules/purefa_pgsched.py
+++ b/plugins/modules/purefa_pgsched.py
@@ -64,7 +64,7 @@ options:
   replicate_frequency:
     description:
     - Specifies the replication frequency in seconds.
-    - Range 900 - 34560000 (FA-405, //M10, //X10i and Cloud Block Store).
+    - Range 900 - 34560000 (FA-405, //M10, //X10 and Cloud Block Store).
     - Range 300 - 34560000 (all other arrays).
     type: int
   snap_at:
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_pgsched.py
```diff
@@ -419,35 +465,124 @@ def update_schedule(module, array, snap_time, repl_time):
...
+                                    all_for_sec=new_repl["target_all_for"],
+                                    per_day=new_repl["target_per_day"],
+                                    days=new_repl["target_days"],
+                                )
+                            ),
+                        )
+                    else:
+                        array.patch_protection_groups(
+                            names=[module.params["name"]],
+                            context_names=[module.params["context"]],
```
skeerthivasan, line 577: context_names should be dropped as it's in else loop

plugins/modules/purefa_pgsched.py
```diff
@@ -305,31 +320,76 @@ def update_schedule(module, array, snap_time, repl_time):
...
+                                )
+                            ),
+                        )
+                    else:
+                        array.patch_protection_groups(
+                            names=[module.params["name"]],
+                            protection_group=ProtectionGroup(
+                                snapshot_schedule=SnapshotSchedule(
+                                    enabled=new_snap["snap_enabled"],
+                                    frequency=new_snap["snap_frequency"],
```
skeerthivasan, line 344: possibly bug in API, whatever value we pass, snap frequency always set to "3600000", pls cross check on the same.

sdodsley: Please recheck this with the latest push. I can't find a direct issue, but it may be an artifact of another issue that has been fixed.

PR #711, opened 2025-02-06
Fix protocol endpoint check to cover all old REST versions

sdodsley:
##### SUMMARY
Change value check to ensure protocol endpoints are correctly identified, even with old REST 2 versions, ie pre-2.16.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_volume.py


skeerthivasan (approved):
Works fine, good to merge.

PR #719, opened 2025-03-03
Fix issue where overwrite not applying correctly

sdodsley:
##### SUMMARY
Fixes issue where a protection group volume is not restoring correctly to a volume group volume, even when `overwrite` is specified.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_pgsnap.py

### DIFF
--- a/plugins/modules/purefa_pgsnap.py
+++ b/plugins/modules/purefa_pgsnap.py
@@ -642,7 +642,7 @@ def restore_pgsnapvolume(module, array):
                         with_default_protection=module.params[
                             "with_default_protection"
                         ],
-                        add_to_protection_group_names=add_to_pgs,
+                        add_to_protection_groups=add_to_pgs,
                         context_names=[module.params["context"]],
                     )
                 else:
@@ -652,7 +652,7 @@ def restore_pgsnapvolume(module, array):
                         with_default_protection=module.params[
                             "with_default_protection"
                         ],
-                        add_to_protection_group_names=add_to_pgs,
+                        add_to_protection_groups=add_to_pgs,
                     )
             else:
                 if LooseVersion(CONTEXT_API_VERSION) <= LooseVersion(api_version):
@@ -670,13 +670,20 @@ def restore_pgsnapvolume(module, array):
                             overwrite=module.params["overwrite"],
                         )
                 else:
-                    res = array.post_volumes(
-                        names=[module.params["target"]],
-                        volume=VolumePost(source=Reference(name=source_volume)),
-                        with_default_protection=module.params[
-                            "with_default_protection"
-                        ],
-                    )
+                    if module.params["overwrite"]:
+                        res = array.post_volumes(
+                            names=[module.params["target"]],
+                            volume=VolumePost(source=Reference(name=source_volume)),
+                            overwrite=module.params["overwrite"],
+                        )
+                    else:
+                        res = array.post_volumes(
+                            names=[module.params["target"]],
+                            volume=VolumePost(source=Reference(name=source_volume)),
+                            with_default_protection=module.params[
+                                "with_default_protection"
+                            ],
+                        )
         else:
             res = array.post_volumes(
                 names=[module.params["target"]],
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_pgsnap.py
```diff
@@ -670,13 +670,23 @@ def restore_pgsnapvolume(module, array):
...
-                        volume=VolumePost(source=Reference(name=source_volume)),
-                        with_default_protection=module.params[
-                            "with_default_protection"
-                        ],
-                    )
+                    if module.params["overwrite"]:
+                        res = array.post_volumes(
+                            names=[module.params["target"]],
+                            volume=VolumePost(source=Reference(name=source_volume)),
+                            with_default_protection=module.params[
```
skeerthivasan, line 677: Cannot pass both variable "overwrite" & "with_default_protection", getting error as invalid parameter combination. We should restrict this in ansible play level as well to refrain users in passing both vars.

bsushanthkumar:
LGTM

plugins/modules/purefa_pgsnap.py
```diff
@@ -917,7 +924,7 @@ def main():
     )
 
     required_if = [("state", "copy", ["suffix", "restore"])]
-    mutually_exclusive = [["now", "remote"]]
+    mutually_exclusive = [["now", "remote"], ["overwrite", "add_to_pgs"]]
```
skeerthivasan, line 927: @sdodsley , Shouldn't we consider adding "with_default_protection" as well.
While am testing in interpreter, passing it also resulted in invalid param combo.

sdodsley: Agreed and done

plugins/modules/purefa_pgsnap.py
```diff
@@ -917,7 +924,11 @@ def main():
     )
 
     required_if = [("state", "copy", ["suffix", "restore"])]
-    mutually_exclusive = [["now", "remote"]]
+    mutually_exclusive = [
+        ["now", "remote"],
+        ["overwrite", "add_to_pgs"],
+        ["overwrite", "with_default_protection"],
```
skeerthivasan, line 930: Below condition check in Line [985](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/719/files#diff-878535a46b810c4e3849ed682a500ba0d124a008cb428c6b8b7f60cc3d36b00eR989) causing errors as "with_default_protection" default value set to true in Line [920](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/719/files#diff-878535a46b810c4e3849ed682a500ba0d124a008cb428c6b8b7f60cc3d36b00eR920). Removing default helped in testing, please check.

sdodsley: Hmm. Given the line 985 check we don't need these exclusives for overwrite... We have to have a default value for `with_default_protection`.

skeerthivasan (commented):
Line [645](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/719/files#diff-878535a46b810c4e3849ed682a500ba0d124a008cb428c6b8b7f60cc3d36b00eR645)  and [655](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/719/files#diff-878535a46b810c4e3849ed682a500ba0d124a008cb428c6b8b7f60cc3d36b00eR655)  should be renamed as **"add_to_protection_groups"** instead of add_to_protection_group_names

plugins/modules/purefa_pgsnap.py
```diff
@@ -144,7 +144,6 @@
     restore: data
     target: data2
     overwrite: true
-    with_default_protection: false
```
skeerthivasan, line 147: Default value for **with_default_protection** is set to true, so if user don't give this option, it will fail with error - "overwrite and add_to_pgs or with_default_protection are incompatable". Better to retain the example so user will set it to false or we can set default value to false.

bsushanthkumar:
LGTM

PR #723, opened 2025-03-07
User module to REST v2 and add timeout and SSH keys

sdodsley:
##### SUMMARY
Update module to use RESTv2
Add new parameters `timeout` and `public_key` to allow API token TTL and SSH public keys for a user.

##### ISSUE TYPE
- Feature Pull Request
- REST 2 Pull Request

##### COMPONENT NAME
purefa_user.py

### DIFF
--- a/plugins/modules/purefa_user.py
+++ b/plugins/modules/purefa_user.py
@@ -39,6 +39,7 @@ options:
     description:
     - Sets the local user's access level to the array
     type: str
+    default: readonly
     choices: [ readonly, ops_admin, storage_admin, array_admin ]
   password:
     description:
@@ -54,6 +55,19 @@ options:
     - Token can be exposed using the I(debug) module
     type: bool
     default: false
+  timeout:
+    description:
+      - The duration of API token validity.
+      - Valid values are weeks (w), days(d), hours(h), minutes(m) and seconds(s).
+    type: str
+    default: "0"
+    version_added: "1.34.0"
+  public_key:
+    description:
+      - Public key for SSH access.
+      - To remove existing key use an empty string
+    type: str
+    version_added: "1.34.0"
 extends_documentation_fragment:
 - purestorage.flasharray.purestorage.fa
 """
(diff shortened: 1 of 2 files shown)


plugins/modules/purefa_user.py
```diff
@@ -103,116 +135,150 @@
...
         if not module.check_mode:
-            try:
-                if not role:
-                    role = "readonly"
-                array.create_admin(
-                    module.params["name"], role=role, password=module.params["password"]
+            res = array.post_admins(
+                names=[module.params["name"]],
+                admin=AdminPost(
+                    role=AdminRole(name=role, password=module.params["password"])
```
skeerthivasan, line 178: **password** should be out of role parameter, its causing issue while user creation.

plugins/modules/purefa_user.py
```diff
@@ -103,116 +135,150 @@
...
-                    try:
-                        array.set_admin(
-                            module.params["name"], role=module.params["role"]
-                        )
-                    except Exception:
+                    res = array.patch_admins(
+                        names=[module.params["name"]],
+                        admin=AdminPatch(role=AdminRole(name=module.params["role"])),
+                    )
+                    if re.status_code != 200:
```
skeerthivasan, line 258: type in **re**, should be res

plugins/modules/purefa_user.py
```diff
@@ -103,116 +135,150 @@
...
-                    if not module.check_mode:
-                        array.delete_api_token(module.params["name"])
-                api_changed = True
-                if not module.check_mode:
-                    user_token["user_api"] = array.create_api_token(
-                        module.params["name"]
-                    )["api_token"]
-            except Exception:
+            ttl = convert_time_to_millisecs(module.params["timeout"])
+            res = array.post_admins_api_tokens(
```
skeerthivasan, line 236: API token should be deleted & created again, delete is missing here

plugins/modules/purefa_user.py
```diff
@@ -103,116 +135,158 @@
...
+                )
+                if res.status_code != 200:
+                    module.fail_json(
+                        msg="Local User {0} password reset failed. Error: {1}"
+                        "Check old password.".format(
+                            module.params["name"], res.errors[0].message
                         )
-            else:
+                    )
+        if module.params["api"]:
```
skeerthivasan, line 235: We've not set api_changed = True in this block, hence ansible reports just ok even though change has happened.
Please check and add it in appropriate place.

plugins/modules/purefa_user.py
```diff
@@ -240,25 +314,24 @@ def main():
             role=dict(
                 type="str",
                 choices=["readonly", "ops_admin", "storage_admin", "array_admin"],
+                default="readonly",
```
skeerthivasan, line 317: For [PR-723](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/723), at line [317](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/723/files#diff-c57bdb52544218dc64b116a395e1ed7a4cb687b7bd7a6cb3b00980f2b473dd47R317), we've added default="readonly"   due to this when we try to recreate just api token(for an existing user) scenario.
User's role getting changed to readonly.
We should change our example at line [121](https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/723/files#diff-c57bdb52544218dc64b116a395e1ed7a4cb687b7bd7a6cb3b00980f2b473dd47R121) to include role, else user's role will be set to default.
I suppose same effect can happen in timeout case as well.

One more observation - lockout_remaining field is not populated in api call output, same in GUI as well
checked directly from interpreter, whereas REST API help doc shows, we should see the field in GET output

skeerthivasan (approved):
LGTM

Issue #724, opened 2025-03-10
Create Volume fails with error: 'Client' object has no attribute 'get_pod'


JoshuaHaus:
**Describe the bug**
We are Creating LUNs on Flasharray for over a year. The last succesfull rtun was in January. Since the last run 5th of march the Playbook fails with the following Traceback.

" warnings.warn(
  Traceback (most recent call last):
    File \"/home/ansible/.ansible/tmp/ansible-tmp-1741614131.821062-205-213449790125452/AnsiballZ_purefa_volume.py\", line 107, in <module>
    _ansiballz_main()
    File \"/home/ansible/.ansible/tmp/ansible-tmp-1741614131.821062-205-213449790125452/AnsiballZ_purefa_volume.py\", line 99, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
    File \"/home/ansible/.ansible/tmp/ansible-tmp-1741614131.821062-205-213449790125452/AnsiballZ_purefa_volume.py\", line 47, in invoke_module
    runpy.run_module(mod_name='ansible_collections.purestorage.flasharray.plugins.modules.purefa_volume', init_globals=dict(_module_fqn='ansible_collections.purestorage.flasharray.plugins.modules.purefa_volume', _modlib_path=modlib_path),
    File \"/usr/lib64/python3.9/runpy.py\", line 225, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
    File \"/usr/lib64/python3.9/runpy.py\", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
    File \"/usr/lib64/python3.9/runpy.py\", line 87, in _run_code
    exec(code, run_globals)
    File \"/tmp/ansible_purestorage.flasharray.purefa_volume_payload_ald7_j7d/ansible_purestorage.flasharray.purefa_volume_payload.zip/ansible_collections/purestorage/flasharray/plugins/modules/purefa_volume.py\", line 1827, in <module>
    File \"/tmp/ansible_purestorage.flasharray.purefa_volume_payload_ald7_j7d/ansible_purestorage.flasharray.purefa_volume_payload.zip/ansible_collections/purestorage/flasharray/plugins/modules/purefa_volume.py\", line 1777, in main
[...]

PR #725, opened 2025-03-10
Create volume in pod fix

sdodsley:
##### SUMMARY
If a volume is to be created in a pod, we are getting a failure as detailed in #724 
This patch resolves that issue.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_volume.py


sdodsley:
@skeerthivasan looks like the QA missed creating volumes in pods. Please get a new QA test added and revalidate this module.

PR #732, opened 2025-03-19
Fix issue with adding/removing existing volume from a PG

### LINKED ISSUE #731: Missing functionality for add/remove volume from PG after volume has already been created.
**Describe the bug**
When updating a volume to add or remove the volume from a protection group, the module returns OK and doesn't perform the addition/removal

**Expected behavior**
Each time a new `add_to_pgs` is called the volume should be added to the named PG and if `state:absent` is provided the volume should be removed from the PG.

Note that SafeModewill come into play for some removal scenarions.

sdodsley:
##### SUMMARY
Add the ability to specify new protection groups in the `add_to_pgs` parameter for an existing volume.
Also add the ability to remove volume from specified protection groups if `state: absent` is set. This will only remove the volume from the specificied protection groups and WILL NOT delete the actual volume.
Closes #731 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_volume.py

### DIFF
--- a/plugins/modules/purefa_volume.py
+++ b/plugins/modules/purefa_volume.py
@@ -1334,6 +1334,45 @@ def update_volume(module, array):
                             module.params["name"], prio_res.errors[0].message
                         )
                     )
+    if module.params["add_to_pgs"]:
+        pgs_now = []
+        if LooseVersion(CONTEXT_API_VERSION) <= LooseVersion(api_version):
+            current_pgs = list(
+                array.get_protection_groups_volumes(
+                    context_names=[module.params["context"]],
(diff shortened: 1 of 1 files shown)


skeerthivasan (commented):
We should add **add_to_pgs** module param at the below line. Without this update is not happening.
https://github.com/Pure-Storage-Ansible/FlashArray-Collection/pull/732/files#diff-01dcd3c8962166f5a5fd653a230bdc0654701a736218d16380cc525966d07fb5R1932

skeerthivasan (approved):

PR #778, opened 2025-07-17
Remove get_system

sdodsley:
##### SUMMARY
`get_system()` function removed as this is only need for REST v1.x calls, and these have all been removed from the Collection.

##### ISSUE TYPE
- REST 2 Pull Request

##### COMPONENT NAME
module_utils/purefa.py


sdodsley:
Merge last after info, network and policy modules are upgraded to REST v2
volume_tags module must be dealt with as well...

PR #783, opened 2025-08-01
Add py314 for devel

sdodsley:
Add Python 3.14 for devel


sdodsley:
Waiting for GitHub to have py3.14 availble

PR #792, opened 2025-08-28
Add context_names support to purefa_ntp

pure-blim:
##### SUMMARY
Add `context` support to `purefa_ntp` so that NTP can be configured on a remote array.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_ntp.py

### DIFF
--- a/changelogs/fragments/792_fusion_ntp.yaml
+++ b/changelogs/fragments/792_fusion_ntp.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - purefa_ntp - Added Fusion support.
--- a/plugins/modules/purefa_ntp.py
+++ b/plugins/modules/purefa_ntp.py
@@ -49,6 +49,14 @@ options:
     - If it is a hex-encoded string, it cannot be longer than 64 characters.
     - Setting this parameter is not idempotent.
     version_added: "1.22.0"
+  context:
+    description:
+    - Name of fleet member on which to perform the ntp operation.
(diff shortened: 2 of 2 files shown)


sdodsley:
@pure-blim couple of things here, which will probably also apply to the other PRs you have submitted.
1. Use `black` as the linter - that is why the Lint test is failing
2. We need a release note to show Fusion support has been added to the module
3. Don't change the version in `galaxy.yaml`. That is done as part of the process for releasing a Collection. Pre-bumping will potentially cause an issue with the Collection release process.
4. Check for trailing whitespaces

pure-blim:
> @pure-blim couple of things here, which will probably also apply to the other PRs you have submitted.
> 
> 1. Use `black` as the linter - that is why the Lint test is failing
> 2. We need a release note to show Fusion support has been added to the module
> 3. Don't change the version in `galaxy.yaml`. That is done as part of the process for releasing a Collection. Pre-bumping will potentially cause an issue with the Collection release process.
> 4. Check for trailing whitespaces

Addressed the comments:
1. Fixed linting issues
2. How do I add a release note?
3. Reverted change to galaxy.yml
4. Trailing whitespaces removed

sdodsley:
@pure-blim here's an example of a note: https://github.com/Pure-Storage-Ansible/FlashArray-Collection/blob/master/changelogs/fragments/707_fusion_pg.yaml. Use the PR number for each

sdodsley (approved):

PR #793, opened 2025-08-28
Add context_names support to purefa_syslog

pure-blim:
##### SUMMARY
Add `context` support to purefa_syslog so that syslog servers can be configured on a remote array.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
purefa_syslog.py


sdodsley (changes requested):
See comments for #792

PR #797, opened 2025-09-09
Fixes to infov2

sdodsley:
##### SUMMARY
Fixes to info v2

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_info.py


ryuyamag:
@sdodsley 
Thank you for this update.
I have confirmed that the current version works correctly in the 6.7.3 environment, just as v1 did.

Regarding the 6.9.x environment, the same "object has no attribute 'items'" error occurred as before. We suspect this might be an issue with py-pure-client (1.74.0).

My customer's environment is 6.7.x, so I think this fix is sufficient.

PR #1000, opened 2026-05-22
Tgroup Ansible Module implementation

pure-blim:
##### SUMMARY
This change adds topology group (tgroup) support to our Ansible Collection.

With the purefa_tgroup module, users will now be able to create, rename, delete tgroups, and to add/remove tgroups or arrays into existing tgroups

purefa_info has also been updated with a "tgroups" subset to list topology group information

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_tgroup


sdodsley:
There is no need to add the new module into the release notes, but you should add it to the reader file that lists all the modules

sdodsley:
@pure-blim you nee to run `black` againt the new code - that is the linting tool we use.

PR #1031, opened 2026-08-06
purefa_policy - handle null password-policy fields and gate max_password_age

### LINKED ISSUE #1030: purefa_policy fails when management password policy lockout_duration is null
## Describe the bug

`everpure.flasharray.purefa_policy` fails when managing the built-in
FlashArray `management` password policy if the API returns
`lockout_duration` as `null`.

The module directly accesses `pwd_policy.lockout_duration`. With the
installed `py-pure-client`, a model field whose value is `null` raises
`AttributeError`, so the module fails before it can compare or apply the
requested policy configuration.

I assume this will also fail for other attributes returning null:

- enforce_dictionary_check
- enforce_username_check
- lockout_duration
- min_character_groups
- min_characters_per_group
- min_password_length

## To Reproduce

1. Use a FlashArray that supports REST API `2.46`.
2. Ensure the existing `management` password policy has `lockout_duration`
   unset.
3. Confirm the API response from:

   ```text
   GET /api/2.46/policies/password?names=management
   ```

   contains:

   ```json
   {
     "lockout_duration": null
   }
   ```

4. Run this task:

   ```yaml
   - name: Configure management password policy
     everpure.flasharray.purefa_policy:
       name: management
       policy: password
       min_password_length: 14
       min_character_groups: 4
       min_characters_per_group: 1
       enforce_dictionary_check: true
       enforce_username_check: true
       password_history: 24
       max_login_attempts: 5
       lockout_duration: 900
   ```

5. Observe the failure:

   ```text
   File ".../purefa_policy.py", line 2007, in update_policy
       "lockout_duration": pwd_policy.lockout_duration,
   AttributeError
   ```

## Expected behavior

A missing or `null` `lockout_duration` should be treated as “not configured.”
The module should then compare it with the requested value and apply the
requested password-policy settings.

For example, the module should safely retrieve the value:

```python
"lockout_duration": getattr(pwd_policy, "lockout_duration", None),
```

The same safe handling may be required for other nullable password-policy
fields accessed directly.

## Screenshots

Not applicable. A sanitized raw JSON response and traceback can be attached if
useful.

## Desktop

- OS: Linux
- Browser: Not applicable
- Version: Not applicable

## Smartphone

Not applicable.

## Additional context

- `everpure.flasharray` collection version: `1.44.0`
- `py-pure-client` version: `1.90.0`
- Python version: `3.14`
- FlashArray REST API version: `2.46`
- The collection's `purefa_policy` module currently declares password-policy
  support from REST API `2.34`; password-policy management is documented as
  introduced in REST API `2.39`.

Do not include API tokens, session tokens, or unredacted array identifiers in
any attached JSON or logs.

sdodsley:
## Summary

Fixes two password-policy problems in `purefa_policy`, both surfaced by #1030.

### 1. `AttributeError` on null password-policy fields (the reported crash)

Updating a password policy read `pwd_policy.<field>` directly for
`enforce_dictionary_check`, `enforce_username_check`, `lockout_duration`,
`min_character_groups`, `min_characters_per_group` and `min_password_length`.

On newer `py-pure-client` the pydantic models raise `AttributeError` when such a
field is `null` — e.g. `lockout_duration` on a `management` policy that has never
had it set — so the module crashed before it could compare or apply the requested
configuration:

```
File ".../purefa_policy.py", line 2007, in update_policy
    "lockout_duration": pwd_policy.lockout_duration,
AttributeError
```

These fields are now read with `getattr(..., None)` (matching the fields already
handled that way), treating `null` as *not configured*: a `None` differs from any
requested value, so the requested setting is still applied, and
`PolicyPassword.to_dict()` omits `None` fields so nothing spurious is sent.

### 2. `max_password_age` was not version-gated

Diffing the `PolicyPassword` model across REST versions shows `max_password_age`
was **added in REST API 2.39** (2.34 introduced password policies; 2.38 added
`context`). The module allowed it whenever password policies were supported
(2.34), so setting it on a 2.34–2.38 array would send an unsupported field.

It is now gated behind `MAX_PASSWORD_AGE_VERSION = "2.39"`, returning a clear
error on older arrays, with a matching note in the module documentation.

`PASSWORD_VERSION` is left at `2.34` — verified against the SDK as the version
that introduced the password-policy endpoints.

## Tests

- `test_update_password_policy_null_fields` — uses a `spec`-limited mock so direct
  attribute access on the nullable fields raises `AttributeError` (mirroring the
  SDK); fails on the old code, passes with the fix.
- `test_max_password_age_requires_2_39` — asserts the new version gate.

Full `test_purefa_policy.py` suite: **128 passed**; `black` clean.

Fixes #1030

### DIFF
--- a/changelogs/fragments/1031_purefa_policy_password_null.yaml
+++ b/changelogs/fragments/1031_purefa_policy_password_null.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+  - purefa_policy - Fix AttributeError updating a password policy when the array returns null for fields such as ``lockout_duration``, by reading them safely and treating null as not configured
+  - purefa_policy - Gate ``max_password_age`` behind FlashArray REST API 2.39, the version that introduced it, so a clear error is returned instead of an unsupported-field failure on older arrays
--- a/plugins/modules/purefa_policy.py
+++ b/plugins/modules/purefa_policy.py
@@ -260,6 +260,7 @@ options:
     - Supported time units are C(w) (weeks), C(d) (days), C(h) (hours), C(m) (minutes), C(s) (seconds).
     - Range between 1 day and 99999 days.
     - A value of 0 or C(0s) disables password expiration.
+    - Requires FlashArray REST API 2.39 or later.
     type: str
     version_added: 1.33.0
   rule_name:
@@ -526,6 +527,7 @@ SECURITY_VERSION = "2.29"
 ABE_VERSION = "2.4"
 PASSWORD_VERSION = "2.34"
 CONTEXT_VERSION = "2.38"
+MAX_PASSWORD_AGE_VERSION = "2.39"
 CA_VERSION = "2.43"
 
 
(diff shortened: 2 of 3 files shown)


Kainim:
Tested successfully against a live FlashArray using py-pure-client 1.93.0. The updated purefa_policy module handled the existing password-policy fields without errors and completed idempotently (changed: false). I also ran the module’s unit tests: all 128 passed. This fix works for our environment.

Kainim (approved):
Tested successfully against a live FlashArray using py-pure-client 1.93.0. The updated purefa_policy module handled the existing password-policy fields without errors and completed idempotently. All 128 purefa_policy unit tests also passed.

PR #1033, opened 2026-08-07
Fix purefa_workload - default an empty context so recommendation-based placement works

### LINKED ISSUE #1032: purefa_workload recommendation based placement causes internal server error
purefa_workload defaults context to `""` and passes it through as `context_names=[""]`, which the Fusion backend rejects with `500 Unidentified internal error`. The error names no cause, so it is very hard to diagnose.

This makes recommendation: true unusable as designed: asking Fusion to choose the placement still requires naming a fleet member by hand, purely to route the call.

Some sibling modules already handle this (e.g.  purefa_hg) and default to a fleet array.

## To Reproduce

Against a fleet member, with no context:

```
- everpure.flasharray.purefa_workload:
    name: my-workload
    preset: my-preset
    recommendation: true
    state: present
```
  
Fails with Recommendation calculation failure ... Unidentified internal error.

**Reproducible outside Ansible**
GET /workloads?names=my-workload&context_names=   → 500 Unidentified internal error
GET /workloads?names=my-workload                  → 400 Workload does not exist.   (truthful)

The same failure hits the non-recommendation path: the module's own first EXAMPLE uses placement: arrayB with no context, and placement is never used for routing only overwritten by the recommendation result.

## Expected behavior

`recommendation: true`
should need no context there is no array to name until Fusion picks one. context_names should never be sent as [""].

## Workaround
Pass any fleet member as `context`; with `recommendation: true` the module overwrites it with the resolved placement.

## Fix
I will raise a PR with a possible fix in a bit

fpfaff:
Fixes #1032.

`context` defaults to `""` and was passed through as `context_names=[""]`, which the Fusion
backend rejects with `500 Unidentified internal error`. This made `recommendation: true` unusable without naming a fleet member by hand — defeating the point of asking Fusion to choose the placement. The documented `placement: arrayB` example failed the same way.

The context is now resolved once in `main()`, before the first API call:

```python
if not module.params["context"]:
    module.params["context"] = (
        module.params["placement"] or list(array.get_arrays().items)[0].name
    )
```

Precedence is explicit `context` > `placement` > local array name. 

Design decisions:

- **No version or fleet guard.** Both are redundant here: the module already fails below API 2.40 and already fails when the array is not a fleet member.
- **`placement` first.** It was previously a documented no-op outside the recommendation path. Using it for routing makes the module's own first EXAMPLE work. The recommendation path is unaffected, it still overwrites both `placement` and `context` with the resolved target before `post_workloads`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_workload

##### ADDITIONAL INFORMATION
Changelog fragment: `changelogs/fragments/1032_purefa_workload_empty_context.yaml`
Also documents the new default on `context` and `placement`.
Four existing `TestMain` cases needed `get_arrays` mocked, since they use a bare `Mock()` array. They failed without it, confirming they exercise the new path. Three cases added covering local-array default, placement default, and explicit context preserved.

```
$ pytest tests/ -q --ignore=collections
1764 passed in 1.11s

$ pytest tests/unit/plugins/modules/test_purefa_workload.py -q
41 passed in 0.04s

$ black --check plugins/modules/purefa_workload.py tests/unit/plugins/modules/test_purefa_workload.py
All done! ✨ 🍰 ✨
2 files would be left unchanged.
```

### DIFF
--- a/changelogs/fragments/1032_purefa_workload_empty_context.yaml
+++ b/changelogs/fragments/1032_purefa_workload_empty_context.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - purefa_workload - Default an empty context to the placement target, or the local array name, so that recommendation-based placement no longer fails with an internal error when no context is named
--- a/plugins/modules/purefa_workload.py
+++ b/plugins/modules/purefa_workload.py
@@ -29,6 +29,8 @@ options:
     - Name of fleet member on which to perform the workload operation.
     - This requires the array receiving the request is a member of a fleet
       and the context name to be a member of the same fleet.
+    - If not specified, defaults to I(placement) when that is set, otherwise
+      to the name of the array receiving the request.
     type: str
     default: ""
   host:
@@ -65,6 +67,9 @@ options:
   placement:
     description:
     - name of target on which the workload will be deployed
+    - Also used as the request context when I(context) is not specified.
+    - Ignored when I(recommendation) is true, as the recommended target
+      replaces it.
     type: str
   recommendation:
     description:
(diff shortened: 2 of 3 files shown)


sdodsley (approved):

sdodsley:
Please sign the CLA and then we can accept this patch

Issue #1034, opened 2026-08-12
purefa_workload: Issues with idempotency, result reporting, wait handling, and check mode


fpfaff:
**Affected:** `everpure.flasharray` 1.44.0 (`purefa_workload`, added 1.33.0)

While automating Fusion workloads against a live fleet in the lab I ran into a cluster of related problems. 
Together they make the module hard to use in a real playbook. 

---

## 1. The module only returns `changed`, so the playbook can't find its own workload

Every path ends in `module.exit_json(changed=changed)`. That's a problem in three concrete ways.

**With `recommendation: true`, Fusion chooses the fleet member — and never tells you.** This is the
one that really hurts, because it's the flagship feature:

```yaml
- name: Create a workload wherever Fusion recommends
  everpure.flasharray.purefa_workload:
    name: app-data
    preset: gold
    recommendation: true
  register: wl

- name: So... which array is it on?
  ansible.builtin.debug:
    msg: "{{ wl }}"     # => {"changed": true}
```

The placement decision is the whole point of asking for a recommendation, and it's discarded. To
find out where the workload landed you have to follow every create with `purefa_info` and search the
fleet by name — and because a workload's identity is `(name, member)` rather than `name`, the same
name can legitimately exist on two members, so that search is ambiguous in exactly the case you
most need it to be right.

**Volume names are generated by the array from the preset.** Anything you'd naturally do next —
snapshot them, add them to a protection group, connect them somewhere else — needs those names, and
there's no way to get them from the task that made them.

**`changed: true` on a delete doesn't tell you what state you're in.** A destroyed workload is still
there, pending eradication, on a retention timer. "Changed" covers both "destroyed, 24h to recover"
and "gone for good", which are very different things to branch on.

**What I'd suggest:** return a `workload` dict on every action that leaves a workload in place —
`name`, `context` (the member, including the one Fusion picked), `status`, `completed`, `destroyed`,
`time_remaining`, and `volumes`. Deliberately *not* a full mirror of the array's own view of the
workload; that's `purefa_info`'s job, and duplicating it just invites the two to disagree.

## 2. Nothing waits for the array, so the next task races it

There's no `wait` option, and several operations are genuinely asynchronous.

Worth being precise here, because it's easy to over-claim: for a simple one-volume preset the create
call does come back synchronously in about 7 seconds already reporting `ready`. So this isn't "the
module always returns too early". It's that whether it returns too early depends on the preset and
[...]

PR #1035, opened 2026-08-12
purefa_workload - fix check mode side effects and unset preset crash

sdodsley:
##### SUMMARY
Addresses part of issue #1034, plus a related traceback found while reviewing it.

Check mode no longer changes the array:

* state=expand created volumes for real in check mode, as the create loop had no check mode guard. The volume configuration is still validated so a bad configuration name fails as before.
* recommendation=true posted a placement recommendation, which creates an object on the array. It is now only requested outside check mode.

The wait for a placement recommendation was an unbounded poll on status != "completed", so a recommendation ending in a failed state hung the task forever. It now fails on a terminal failure status, gives up after RECOMMENDATION_TIMEOUT seconds, and checks the response of each poll instead of indexing into it blindly.

preset was unconditionally prefixed with the fleet name, so every state that does not need a preset - delete, eradicate, recover and rename - raised a TypeError when it was not supplied. It is now only prefixed when supplied, required_if requires it for state=expand, and the create path reports a clear error when it is missing.

Also corrects the rename example, which used a non-existent state=rename.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
purefa_workload.py


sdodsley:
Closing this as superseded rather than resolving the conflicts.

`purefa_workload.py` was rewritten on master by #1038 and #1034 after this branched, and that work absorbed each of the fixes here - in most cases in a better form. Master's own `changelogs/fragments/1034_purefa_workload_return_facts.yaml` already announces them:

| This PR | On master |
| --- | --- |
| `state=expand` created volumes for real in check mode | Guarded in `expand_workload()` and again inside `_create_volume()`. Fragment: *"Stop `state=expand` from creating volumes for real while running in check mode"* |
| `host` connected in check mode during an expand | `_connect_host()` guards its own `post_connections`, so the call is check-mode safe wherever it is reached from |
| Unbounded poll on the placement recommendation | `_wait_for_recommendation()` uses the shared `wait_for()` helper added by #1034, with an `is_failed` hook for the failed status, a timeout taken from `wait_timeout` rather than a hardcoded 300s, `check_response` on every poll, and a guarded walk down to the target name |
| `preset` unconditionally prefixed with the fleet name | `_read_preset()` prefixes it only on the paths that need a preset. Fragment: *"`preset` is only read when creating or expanding a workload, so omitting it on a delete, rename, recover, or host operation no longer raises a TypeError"* |
| `required_if` should require `preset` for `state=expand` | Already in `required_if`. Fragment: *"Require `preset` for `state=expand`, and fail with a message naming the option"* |
| Rename example used a non-existent `state=rename` | Master drops the `state` line from that example entirely, which is right since `present` is the default |

Verified by merging master into this branch in a scratch worktree: 7 conflict hunks in the module and 6 in the tests, and master's side is the superset on every one. Resolved in master's favour, the branch adds exactly one file over master - its own changelog fragment, which would double-report fixes master already announces. The test file here is 1558 lines against master's 6753, which covers every check-mode path plus `TestReadingThePresetForTheAction` and `TestCheckingOptionCombinations`.

One item was a genuine difference rather than a conflict: this PR stops check mode requesting a placement recommendation, on the grounds that the POST creates an object on the array, while master deliberately keeps it -

```python
# A calculation, not a change, so this runs in check mode too - it is what lets a
# check-mode run report the target Fusion would have chosen
```

Master's position stands, so nothing is being carried over. The branch is left in place in case any of it is wanted later.

PR #1038, opened 2026-08-16
purefa_workload - return workload and fix series of issues

### LINKED ISSUE #1034: purefa_workload: Issues with idempotency, result reporting, wait handling, and check mode

fpfaff:
## Make `purefa_workload` usable from a real playbook

Fixes #1034.

The implementation I promised in the issue. It closes everything I reported, plus the smaller
things that turned up alongside. I wanted to break it down into a set of smaller PRs, however the
nature of the changes made that hard, so I decided to go with one big PR.

It's stacked on #1032, which it supersedes with better logic, so that fragment is deleted here and
its line folded into `1034_purefa_workload_return_facts.yaml`.

### 1. A task now tells you what it did

Every action now returns a `workload` dict with `name`, `context`, `status`, `completed`,
`destroyed`, `time_remaining`, `volumes`. Deliberately not a mirror of the array's own properties.
That stays `purefa_info`'s job.

- **`recommendation` reports where Fusion put it.** `context` is the member Fusion chose. This also
  works for checks, although there is no guarantee that Fusion will pick the same array next time.
- **Volume names come back.**
- **A delete says which state you are in.** `destroyed: true` plus the `time_remaining` countdown.
  An eradicate has nothing left to describe, so it reports `name` and `context` alone — which
  workload, and the array it went from. The dict is empty only when there was no workload at all.

`completed` is derived from `status`, so the two cannot disagree, and an unrecognised status counts
as still in flight rather than as success.

### 2. `wait` and `wait_timeout`, on by default

The polling itself is a new generic `wait_for` helper in `module_utils/api_helpers.py`
(backoff + timeout), so other modules with async operations can use it.

Four things get waited on: the placement calculation, the workload reaching its status, the host
connections, and the volumes an expand adds. Check mode skips the waiting.

Worth noting:

- The timeout bounds **each** operation, not the task. A create with `recommendation` *and* `host`
  waits three times and can take three times `wait_timeout`. That is documented on the option.

### 3. `--check` no longer provisions storage

`state: expand` skips the volume creation in check mode, guarded both in the loop and inside
`_create_volume`, so no future caller can create one by forgetting.

### 4. Re-runs are now idempotent and safe

Both causes, handled differently:

- **The `recommendation` re-run is fixed.** The fleet is searched *before* Fusion is asked, so a
  second run finds what the first placed and converges instead of asking for a new placement.
- **`context`/`placement` now default to the whole fleet** rather than to the array receiving the
  request. A task naming neither finds the workload wherever it is. Creates must either use
  `recommendation: true` or specify `context`/`placement`, as there is nothing to look up yet.
- If there are **multiple possible matches** for `context`/`placement`, the module will **fail**
  stating it cannot determine what to work with.
- **You can have the same workload on multiple arrays**, if you specify the array member explicitly
  via `context`. This is deliberate: the identity of a workload is `(name + fleet member)`, so it is
  a second workload, not a duplicate, and the array permits it. It is no longer silent: the create
  warns naming the other members. Works for all operations.

### `main()` got reworked

`main()` is now four steps, making it easier to maintain and test the module.

1. are the options coherent
2. which array do we address
3. what is there
4. what to do about it

The last is a pure function returning a decision, which is what allows us to test the whole
decision logic without mocking an array.

### Other issues addressed

- `placement` was ignored outside a `recommendation`-based create; it now takes effect everywhere.
  It stays a separate option from `context` rather than an alias, so naming both with different
  members is refused instead of one silently winning.
- Options that were accepted and did nothing are now warned about: `volume_count`,
  `volume_configuration` and `parameters`. The warning names the action the task resolved to.
- Actions that are impossible are now refused, e.g. `eradicate` outside `state: absent`, or `host`
  with `rename`.
- Crashes became messages: a missing workload, an unlistable volume set, or a placement Fusion
  cannot find now fail with something you can act on. Every SDK field is optional and raises on
  null, so the reads are guarded.
- Absence is read from the array's error message rather than just the error code — the endpoint
  answers 400 for several distinct issues and only one means "not there".
- `preset` is only read where it is used, so omitting it on a delete, rename, recover or host task
  no longer raises a `TypeError`; an expand that omits it is told which option is missing.
- Impossible renames no longer create or recover a workload, and `rename` with
  `state: absent`/`expand` is refused rather than ignored.
- `state: expand` against a missing or destroyed workload fails instead of reporting no change.
- Connecting a `host` where some volumes are already attached no longer errors, and `state: absent`
  with both `host` and `eradicate` no longer eradicates — naming a host makes it a disconnect.
- `volume_count: 0` is rejected again (regression from 1.0.34).
- A delete or eradicate that resolved its own target says which member it chose, first.

### Breaking changes

Three potentially breaking changes:

- the fleet-wide `context`/`placement` default
- the refusal when two members hold the name
- `eradicate`/`host` no longer accepted where they do nothing

### Testing

Unit tests went 41 → 304, and the module was exercised against a live Fusion fleet with a
verification playbook covering each phase in check mode and for real, including the
recommendation-based deploys and a repeat run to prove convergence.


fpfaff:
@sdodsley do we really have to be compatible with python 2.7 and 3.6 ?
3.6 is EOL since 2021. 

If we really have to I can rewrite the dataclass as a class. But it would be a shame to not be able use a python concept that exists since ~2018.

sdodsley:
@fpfaff no - i'll fix it - its a new thing from Red Hat for the certification we have to do.

PR #1045, opened 2026-09-03
feat(purefa_timezone): Add module to manage the array system time zone

### LINKED ISSUE #1044: Feature request: module to manage the array time zone (purefa_timezone)
**Is your feature request related to a problem? Please describe.**

There is no module in the collection to manage the array's system time zone. The collection covers the other "day-0 baseline" settings we enforce fleet-wide (`purefa_dns`, `purefa_ntp`, `purefa_ds`/`purefa_dsrole`, `purefa_smtp`, `purefa_alert`, `purefa_banner`, `purefa_timeout`, `purefa_eradication`), but time zone is the one setting we still have to reach around the collection for.

Either:

1. A new module, e.g. `purefa_timezone`, mirroring the shape of `purefa_timeout`:

   ```yaml
   - name: Set array time zone
     everpure.flasharray.purefa_timezone:
       timezone: America/New_York
       fa_url: "{{ fa_url }}"
       api_token: "{{ api_token }}"
   ```

   Validation of the value against the IANA tz database (the collection already lists `pytz` as a requirement) and reporting via `purefa_info` (`default` subset) would round it out.

2. Or, add a `timezone` option to an existing array-level module such as `purefa_arrayname` (renamed/generalised) — though a dedicated module seems cleaner and more consistent with how the other array-level settings are split.

The underlying API is already there: Purity REST 2.x exposes `time_zone` on `GET /arrays` and accepts it on `PATCH /arrays`. From our testing on Purity 6.x arrays the field appears from REST **2.26** onward (it is absent in 2.17), and works through 2.46. With py-pure-client that's `arrays.get_arrays()` / `arrays.patch_arrays(array=Arrays(time_zone=...))`.

**Describe alternatives you've considered**

We currently work around it with `ansible.builtin.uri` against the REST API directly (login → `GET /api/2.26/arrays` → compare → `PATCH` only if different → logout). It works and is idempotent, but it means:

- hand-rolling session/token handling that every other task in the play gets from the collection for free,
- pinning a REST version separately from what the collection negotiates,
- `no_log` gymnastics because the raw session token passes through `uri`,
- no `--diff` output without extra debug tasks.

A first-class module would remove all of that.

**Additional context**

- Collection version tested: 1.45.0 (`purestorage.flasharray` namespace, via the `everpure.flasharray` redirects)
- Purity//FA 6.x; REST API versions confirmed: `time_zone` absent in 2.17, present in 2.26–2.46

sdodsley:
##### SUMMARY
The collection covers the other day-0 baseline settings - DNS, NTP, directory services, SMTP, alerts, banner, GUI timeout and eradication - but the system time zone had to be reached around the collection for, typically with ansible.builtin.uri against the REST API directly.

The new module sets the array time zone, taking an IANA time zone name. The value is validated against the pytz time zone database, which the collection already requires, before the array is contacted, and the module requires REST API 2.26 or higher, which is where time_zone appears on the arrays endpoint. Fleet members are supported through the usual context parameter.

The SDK flags time_zone as a read-only field of the Arrays model, and request bodies are serialised with as_request_dict(), which drops every read-only field. A body built as Arrays(time_zone=...) therefore serialises to {} and patches nothing at all, while the module would still report changed. The body is instead passed as a dict, which the serialiser sends verbatim, and because the field is read-only by the spec, the time zone the array echoes back in its response is compared against the requested one so that a quietly ignored field is reported as a failure rather than as success.

The current time zone of an array is already reported by the config subset of purefa_info.

Closes #1044

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
purefa_timezone.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ All modules are idempotent with the exception of modules that change or set pass
 - purefa_tgroup - manage topology groups and their memberships on the FlashArray
 - purefa_token - manage FlashArray user API tokens
 - purefa_timeout - manage the GUI idle timeout on the FlashArray
+- purefa_timezone - manage the system time zone on the FlashArray
 - purefa_user - manage local user accounts on the FlashArray
 - purefa_vg - manage volume groups on the FlashArray
 - purefa_vlan - manage VLAN interfaces on the FlashArray
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -63,6 +63,7 @@ action_groups:
    - purefa_syslog_settings
    - purefa_tgroup
    - purefa_timeout
+   - purefa_timezone
    - purefa_token
    - purefa_user
    - purefa_vg
--- a/plugins/modules/purefa_timezone.py
+++ b/plugins/modules/purefa_timezone.py
@@ -1,0 +1,162 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2026, Simon Dodsley (simon@everpuredata.com)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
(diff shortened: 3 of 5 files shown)


sdodsley:
There is no POST capable way to set the timezone through the API. Thi scan only be done through support.

PR #1057, opened 2026-09-09
update purefa_inventory to retain drive serial, interface pcie slot, and move temp_sensors to the proper dict

meagan-gibbons-exp:
##### SUMMARY
`purefa_inventory` did not expose two pieces of hardware data that the REST API makes available, which forced our NetBox hardware-sync workflows to bypass the collection and call the raw API directly. This exposes both, and (via a breaking change) fixes a related issue with temp sensors that irks me.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module: purefa_inventory

##### ADDITIONAL INFORMATION
1. The slot on which an interface or other module lives is provided by the hardware API but isn't returned by `purefa_inventory`. This is useful for syncing our FlashArray hardware data to netbox to correlate specific interfaces to the pcie expansion cards on which they reside. I would also add slot to the direct_compress_accelerator card for the same reason (tracking which pcie slot it's in), but since the DCA is not returned by `purefa_inventory` today–and doesn't fit neatly into one of the existing `hw_info` dicts–I wasn't sure where to place it, so I left it out for now. I'll have to keep consuming the API directly for tracking the DCA unless someone has thoughts about where to place it in the `purefa_inventory` module's output, but that's a "me" problem until I get some input there on where it should go.

2. The drives section contains what appears to be an unintentional regression where the `status`, `identify_enabled`, and `serial` attributes are written into the `hw_info` dict but then discarded by the `for drive in drives` pass setting a dict literal in the same location. The PR fixes this by doing a merge instead; also, it adds the `nvram_bay` to the component pass. The `nvram_bay`s are returned by the `get_drives` pass but if we don't add it to the components pass in conjunction with the merge fix for the drives issue, then those items would still be missing their `identify_enabled` attribute (as they do today). The merge defaults to None for the `identify_enabled` and `serial` in case `get_drives` ever returns another type not accounted for in the components pass in a similar fashion to the `nvram_bay` issue I just described.

3. Temperature readings are moved into the already existing `hw_info['temperature']` dict instead of the `hw_info['controllers']` dict. As it stands today the temperatures dict is returned by purefa_inventory but is always empty and the temp sensors are all in controllers instead. Not sure why that is but this should fix that. Note that **this would be a breaking change** for anyone's code that was expecting the temperature readings in the `controllers` dict, though, so please let me know if I should divorce this change from this PR to prevent introducing a breaking change the collection isn't ready for.

I had `claude` help with the updates to the module's tests covering the changes.

[...]

PR #1059, opened 2026-09-10
purefa_inventory - report unclassified hardware in an other dict

### LINKED ISSUE #1058: purefa_inventory does not return the direct_compress_accelerator
**Describe the bug**
`purefa_inventory` does not return the `direct_compress_accelerator`. This information is returned by the /hardware API but not included in the `purefa_inv` response

**To Reproduce**
Run `purefa_inventory` on an array with installed `direct_compress_accelerator` pcie expansion cards. 

**Expected behavior**
`purefa_inventory` returns all the hardware inventory components in the array, not a subset of them.

**Code Snip**
/hardware returns the component
```
[pit2-pure-05] TASK: gather_data : fail (debug)> p task_vars['fa_hardware_raw']
{'changed': False,
 'connection': 'close',
 'content_security_policy': "frame-ancestors 'none'",
 'content_type': 'application/json',
 'cookies': {},
 'cookies_string': '',
 'date': 'Thu, 10 Sep 2026 12:27:58 GMT',
 'elapsed': 0,
 'failed': False,
 'json': {'continuation_token': None,
          'items': [{...},
...
                    {'details': None,
                     'identify_enabled': None,
                     'index': 4,
                     'model': None,
                     'name': 'CT0.DCA4',
                     'serial': None,
                     'slot': 4,
                     'speed': None,
                     'status': 'ok',
                     'temperature': None,
                     'type': 'direct_compress_accelerator',
                     'voltage': None},
...
```
`purefa_inventory` does not return the DCA
```
[pit2-pure-05] TASK: gather_data : fail (debug)> p task_vars['purefa_inventory']['purefa_inv']
{'chassis': {...},
 'controllers': {...},
 'drives': {...},
 'fans': {...},
 'interfaces': {...},
 'power': {...},
 'temperature': {...}}
```

**Desktop (please complete the following information):**
 - OS: Ubuntu 26.04 LTS
 - Browser n.a
 - Version everpure.flasharray v1.44.0

**Additional context**
I referenced this in my recent PR adding the slot number to the interfaces, and figured I'd raise it as a bug report. Adjusting the module to return the DCA and other non-NIC pcie expansion cards *should* be a fairly simple change, but I didn't include it in that PR since the current shape of purefa_inv doesn't include a section into which the DCA fits neatly, eg.,
```
hw_info = {
        "fans": {},
        "controllers": {},
        "drives": {},
        "interfaces": {},
        "power": {},
        "chassis": {},
        "temperature": {},
    }
```
We could just add an 'other' dict to catch anything not included in the existing dicts, but if there's somewhere else this could/should land then happy to go that direction as well.

sdodsley:
##### SUMMARY

Fixes #1058

`generate_new_hardware_dict()` only reported the component types it had a dedicated dict for, so anything else `/hardware` returned was silently dropped. That is `direct_compress_accelerator`, which prompted the report, and also `sas_module` and `storage_shelf` - both in the documented type list, neither with a home.

Unclassified components now go into a new `other` dict, keyed by component name and carrying the component `type`, so new hardware types get reported as arrays start returning them rather than needing a module change each time. The chain of independent `if` statements becomes an `if`/`elif`/`else` so the `else` can catch them; the conditions were already mutually exclusive, so nothing about the existing dicts changes, and it no longer tests every component against all seven branches.

Example, for the DCA from the issue:

```yaml
other:
  CT0.DCA4:
    type: direct_compress_accelerator
    status: ok
    slot: 4
    index: 4
    model: null
    serial: null
    identify_enabled: null
    speed: null
    temperature: null
    voltage: null
    details: null
```

**Why a catch-all rather than a dedicated dict per type.** `nvram_bay` is handled by this module but is not in the API's documented type list, so enumerating types will keep losing this race. A catch-all fixes the whole class at once, which is what the issue asks for - all the hardware components, not a subset.

To keep that from becoming a compatibility trap, `other` is documented as unstable in both the module `description` and the changelog: a component reported there may move into a dedicated dict in a future release, and consumers should match on the `type` key. That makes any such promotion a minor change rather than a breaking one.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_inventory

##### ADDITIONAL INFORMATION

**No REST version gate is needed.** `direct_compress_accelerator` becomes a documented type value at REST 2.51 - the SDK carries two generations of the `Hardware` model and that is where the type list flips from the old short forms (`bay`, `ct`, `eth`, `fc`, `tmp`...) to the long ones. But `type` is an unvalidated `Optional[StrictStr]` in both generations, with no enum validator, so the value deserialises at any REST version and the new branch is simply inert on an array that does not report such components. `slot` is likewise present in `hardware_v_2_2`, so #1057 needed no gate either.

**Every field in `other` is read with a `getattr` default.** Which fields a component populates is not knowable in advance, and the SDK models raise `AttributeError` rather than returning `None` for the ones it does not. From the payload in the issue, a DCA populates only `status`, `slot` and `index` - so the bare `component.identify_enabled` pattern the `drive_bay` branch uses would crash on one. That is the trap this fix has to avoid, and there is a test for it.

**Testing.** The DCA test uses the exact payload from the issue, via a `FakeComponent` stand-in that raises `AttributeError` on null fields the way the SDK models do. A plain `Mock` returns `None` instead, which would hide the crash above.

Verified the new tests bite, by reverting the module change and re-running them:

```paste below
FAILED tests/unit/plugins/modules/test_purefa_inventory.py::TestOtherHardware::test_reports_direct_compress_accelerator
FAILED tests/unit/plugins/modules/test_purefa_inventory.py::TestOtherHardware::test_reports_unknown_component_type
FAILED tests/unit/plugins/modules/test_purefa_inventory.py::TestOtherHardware::test_known_component_types_stay_out_of_other
FAILED tests/unit/plugins/modules/test_purefa_inventory.py::TestOtherHardware::test_other_is_always_present
4 failed, 13 deselected
```

Full suite after the fix: `pytest tests/unit` - 2463 passed. `black --check` clean, `yamllint -c .yamllint` clean on the fragment, and `DOCUMENTATION`/`EXAMPLES`/`RETURN` all still parse as YAML after the doc edit.

**Not verified against hardware.** I enumerated a lab array at REST 2.54, which reports 124 components across 9 types - `chassis`, `controller`, `cooling`, `drive_bay`, `eth_port`, `fc_port`, `nvram_bay`, `power_supply`, `temp_sensor` - and confirmed `other` stays empty for all of them. It has no DCA, `sas_module` or `storage_shelf` cards, so those three paths are covered by unit tests only. @meagan-gibbons-exp - if you get a chance, it would be worth confirming the `other` output on the array from the issue before this ships.

Changelog fragment included: `changelogs/fragments/1059_purefa_inventory_other_hardware.yaml`.

### DIFF
--- a/changelogs/fragments/1059_purefa_inventory_other_hardware.yaml
+++ b/changelogs/fragments/1059_purefa_inventory_other_hardware.yaml
@@ -1,0 +1,8 @@
+minor_changes:
+  - purefa_inventory - Hardware components without a dedicated dict of their own are now reported in
+    a new ``other`` dict, keyed by component name and carrying the component ``type``, rather than
+    being dropped from the inventory. This covers ``direct_compress_accelerator``, ``sas_module``
+    and ``storage_shelf``, as well as any component type the array reports in future.
+  - purefa_inventory - The contents of the ``other`` dict are not stable. A component reported there
+    may be moved into a dedicated dict in a future release, so match on the ``type`` key rather than
+    assuming a component stays in ``other``.
--- a/plugins/modules/purefa_inventory.py
+++ b/plugins/modules/purefa_inventory.py
@@ -21,6 +21,12 @@ short_description: Collect information from Everpure FlashArray
 version_added: '1.0.0'
 description:
   - Collect hardware inventory information from a Everpure Flasharray
+  - Hardware components that do not have a dedicated dict of their own are
+    reported in the C(other) dict, keyed by component name, with the component
+    type in the C(type) key.
+  - The contents of C(other) are not stable. A component reported there may be
+    moved into a dedicated dict in a future release, so match on the C(type)
+    key rather than assuming a component stays in C(other).
 author:
   - Everpure ansible Team (@sdodsley) <pure-ansible-team@everpuredata.com>
 extends_documentation_fragment:
@@ -66,6 +72,7 @@ def generate_new_hardware_dict(array):
         "power": {},
         "chassis": {},
         "temperature": {},
+        "other": {},
     }
     components = list(array.get_hardware().items)
     for component in components:
@@ -77,18 +84,18 @@ def generate_new_hardware_dict(array):
                 "model": component.model,
                 "identify_enabled": component.identify_enabled,
             }
-        if component.type == "controller":
+        elif component.type == "controller":
             hw_info["controllers"][component_name] = {
                 "status": component.status,
                 "serial": component.serial,
                 "model": component.model,
                 "identify_enabled": component.identify_enabled,
             }
-        if component.type == "cooling":
+        elif component.type == "cooling":
             hw_info["fans"][component_name] = {
                 "status": component.status,
             }
-        if component.type == "temp_sensor":
+        elif component.type == "temp_sensor":
             sensor = {
                 "status": component.status,
                 "temperature": component.temperature,
@@ -99,7 +106,7 @@ def generate_new_hardware_dict(array):
             # compatible addition. Deprecated - to be removed from
             # ``controllers`` in the next major release.
             hw_info["controllers"][component_name] = dict(sensor)
-        if component.type in [
+        elif component.type in [
             "drive_bay",
             "nvram_bay",
         ]:
@@ -108,7 +115,7 @@ def generate_new_hardware_dict(array):
                 "identify_enabled": component.identify_enabled,
                 "serial": getattr(component, "serial", None),
             }
-        if component.type in [
+        elif component.type in [
             "sas_port",
             "fc_port",
             "eth_port",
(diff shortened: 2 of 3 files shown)


meagan-gibbons-exp:
seems to work

here's an array that has a shelf and DCA v2's
```
[cmh2-pure-01] TASK: gather_data : fail (debug)> p task_vars['purefa_inventory']['purefa_inv']
...
 'other': {'CT0.DCA4': {'details': None,
                        'identify_enabled': None,
                        'index': 4,
                        'model': 'v2',
                        'serial': None,
                        'slot': 4,
                        'speed': None,
                        'status': 'ok',
                        'temperature': None,
                        'type': 'direct_compress_accelerator',
                        'voltage': None},
           'CT1.DCA4': {'details': None,
                        'identify_enabled': None,
                        'index': 4,
                        'model': 'v2',
                        'serial': None,
                        'slot': 4,
                        'speed': None,
                        'status': 'ok',
                        'temperature': None,
                        'type': 'direct_compress_accelerator',
                        'voltage': None},
           'SH9': {'details': None,
                   'identify_enabled': False,
                   'index': 9,
                   'model': 'DFS2',
                   'serial': 'SERIAL_NUMBER',
                   'slot': None,
                   'speed': None,
                   'status': 'ok',
                   'temperature': 26,
                   'type': 'storage_shelf',
                   'voltage': None},
           'SH9.SC0': {'details': None,
                       'identify_enabled': False,
                       'index': 0,
                       'model': 'DFSC2',
                       'serial': 'SERIAL_NUMBER',
                       'slot': None,
                       'speed': None,
                       'status': 'ok',
                       'temperature': 32,
                       'type': 'shelf_controller',
                       'voltage': None},
           'SH9.SC1': {'details': None,
                       'identify_enabled': False,
                       'index': 1,
                       'model': 'DFSC2',
                       'serial': 'SERIAL_NUMBER',
                       'slot': None,
                       'speed': None,
                       'status': 'ok',
                       'temperature': 32,
                       'type': 'shelf_controller',
                       'voltage': None}},
...
```
And from the pit2-pure-05 array from the original issue which just has dca version 1's:
```
[pit2-pure-05] TASK: gather_data : fail (debug)> p task_vars['purefa_inventory']['purefa_inv']
...
 'other': {'CT0.DCA4': {'details': None,
                        'identify_enabled': None,
                        'index': 4,
                        'model': None,
                        'serial': None,
                        'slot': 4,
                        'speed': None,
                        'status': 'ok',
                        'temperature': None,
                        'type': 'direct_compress_accelerator',
                        'voltage': None},
           'CT1.DCA4': {'details': None,
                        'identify_enabled': None,
                        'index': 4,
                        'model': None,
                        'serial': None,
                        'slot': 4,
                        'speed': None,
                        'status': 'ok',
                        'temperature': None,
                        'type': 'direct_compress_accelerator',
                        'voltage': None}},
...
```

meagan-gibbons-exp:
Thanks!

PR #1061, opened 2026-09-14
purefa_user - allow AD usernames that local account rules reject

### LINKED ISSUE #1060: Adding a public key for an AD user fails if the AD username doesn't meet purity username requirements
**Describe the bug**
FlashArray using LDAP authentication where usernames are in the format of `first.last`, which does not meet purity's requirements for usernames `name must contain a minimum of 1 and a maximum of 32 characters (alphanumeric or `-`). All letters must be lowercase.`. When attempting to assign a public key to the user via the `everpure.flasharray.purefa_user`, an error regarding the format of the username is received. 

**To Reproduce**
```
- name: Assign the provided SSH public key to an AD user
  gather_facts: false
  hosts: pure
  vars_files:
    - tokens/{{ inventory_hostname }}.yml
  vars_prompt:
    - name: username
      prompt: "Please enter the username:"
      private: false
    - name: public_key_file_path
      prompt: "Please enter the full path to a file containing the public key:"
      private: false
  tasks:
    - name: Sync public key to array user.
      everpure.flasharray.purefa_user:
        name: "{{ username }}"
        ad_user: true
        public_key: "{{ lookup('file', public_key_file_path) }}"
        fa_url: "{{ mgmt_vip }}"
        api_token: "{{ token }}"
      delegate_to: localhost
```
```
meagan@EXP20237:~/git-repos/ansible-pure-storage$ ansible-playbook ~/git-repos/ansible-pure-storage/assign_ssh_pubkey_ad_user.yml --diff --limit 'pit2-pure-02'
[WARNING]: skipping vars_files item due to an undefined variable
Origin: /home/meagan/git-repos/ansible-pure-storage/assign_ssh_pubkey_ad_user.yml:6:7

4   hosts: pure
5   vars_files:
6     - tokens/{{ inventory_hostname }}.yml
        ^ column 7

Please enter the username:: meagan.gibbons
Please enter the full path to a file containing the public key:: /mnt/c/Users/meagan.gibbons/.ssh/id_ed25519.pub

PLAY [Assign the provided SSH public key to an AD user] ****************************************************************************************************************************************************

TASK [Sync public key to array user.] **********************************************************************************************************************************************************************
[ERROR]: Task failed: Module failed: name must contain a minimum of 1 and a maximum of 32 characters (alphanumeric or `-`). All letters must be lowercase.
Origin: /home/meagan/git-repos/ansible-pure-storage/assign_ssh_pubkey_ad_user.yml:15:7

13       private: false
14   tasks:
15     - name: Sync public key to array user.
         ^ column 7

fatal: [pit2-pure-02 -> localhost]: FAILED! => {"changed": false, "msg": "name must contain a minimum of 1 and a maximum of 32 characters (alphanumeric or `-`). All letters must be lowercase."}

PLAY RECAP *************************************************************************************************************************************************************************************************
pit2-pure-02               : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

**Expected behavior**
The public key is assigned to the user.

**Screenshots**

Image

Image

**Desktop (please complete the following information):**
 - OS: Ubuntu 26.04
 - Browser ansible [core 2.21.1]
 - Version everpure.flasharray.purefa_user 1.44.0

**Additional context**

There is no issue with the format of the LDAP usernames for logging in via GUI or SSH, and public keys can be assigned to the LDAP user via the GUI or the Purity CLI.

sdodsley:
##### SUMMARY

Fixes #1060

The module applied Purity's **local** account naming rules - lowercase alphanumeric and hyphen, 32 characters - to every name, before making any API call:

```python
pattern = re.compile("^[a-z0-9]([a-z0-9-]{0,30}[a-z0-9])?$")
if not pattern.match(module.params["name"]):
    module.fail_json(msg="name must contain a minimum of 1 and a maximum of 32 characters ...")
```

Directory service users are named by the directory, not by Purity, so an LDAP user such as `first.last` was rejected by us and the array never saw the request. The GUI and the Purity CLI both accept these names, as the reporter notes.

The pattern predates AD support - it came in with the original local-user module. When `ad_user` was added, the new `update_ad_user()` and `delete_ad_user()` paths were written but the validation above them was never made conditional. It is now applied only to local users, which also unblocks `state: absent` for AD users, blocked the same way today.

**Confirmed against an array at REST 2.54:**

```
get_admins(names=['no.such.user'])                      -> 400: Unable to find specified user.
patch_admins(['zz-no-such.user-probe'], public_key="")  -> 200
patch_admins(['zznosuchuserprobe'],     public_key="")  -> 200   # plain-name control
```

The 400 is an existence error, not a format error - the API parsed the dotted name. The PATCH the module actually makes accepts a dotted name outright. Both probe names were chosen so they could not exist, and the admin count was unchanged afterwards, so nothing was modified.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
purefa_user

##### ADDITIONAL INFORMATION

**Three further defects in the same module, found while verifying the above.**

**1. Check mode wrote to the array.** The module declares `supports_check_mode=True`, but `update_ad_user()` had no check mode guard at all, and `create_local_user()` guarded its password and role patches but not the API token block or the public key patch. So `ansible-playbook --check` against an existing user *revoked and reissued a live API token*, breaking whatever held the old one, and overwrote the SSH key. The existing tests covered check mode for user creation and deletion but neither update path, which is why it went unnoticed.

**2. An empty `public_key` did not clear an AD user's key.** The documentation says "To remove existing key use an empty string", and `create_local_user` correctly tests the parameter against `None`, but `update_ad_user` tested it for truthiness - so the empty string was skipped and the task reported no change having done nothing.

**3. Assigning a key to an AD user always reported `changed`.** It is now compared against the current key where the array has one to compare against. An AD user with no array-side state is not returned by `get_admins`, so there is nothing to compare in that case and the patch goes ahead.

**Deliberately not included.** `patch_admins` returns 200 for a user that does not exist, so a misspelled AD username reports `changed=True` and the key silently goes nowhere. An existence check cannot fix this: an unprovisioned directory user and a typo are indistinguishable on the way in, since `get_admins` does not return directory users with no array-side state - both give the same 400. A post-write check looks viable, because the bogus-name probe above left no array-side state while a real directory user would gain some, but that rests on array behaviour I could not verify without provisioning an AD user. Worth a separate issue.

**Three existing tests needed `check_mode = False` added.** They construct `Mock()` without setting it, so `check_mode` was a truthy `Mock` - harmless while the function ignored check mode entirely, but now it means "check mode". They are testing real runs, so the attribute is now set explicitly. That they failed is the guard working.

Verified the new tests bite, by reverting the module change and re-running them:

```paste below
FAILED ...::TestAdUserNameValidation::test_ad_username_with_dot_is_accepted
FAILED ...::TestUpdateAdUserCheckMode::test_api_token_not_recreated_in_check_mode
FAILED ...::TestUpdateAdUserCheckMode::test_public_key_not_written_in_check_mode
FAILED ...::TestCreateLocalUserCheckMode::test_api_token_not_recreated_in_check_mode
FAILED ...::TestCreateLocalUserCheckMode::test_public_key_not_written_in_check_mode
FAILED ...::TestAdUserPublicKey::test_empty_public_key_clears_existing_key
FAILED ...::TestAdUserPublicKey::test_matching_public_key_reports_no_change
7 failed, 3 passed
```

The three that pass either way are the guards against over-correcting: a local user named `first.last` must still be rejected, an uppercase local name must still be rejected, and clearing a key on a user with no array-side state must still report no change.

Full suite after the fix: `pytest tests/unit` - 2473 passed. `black --check` clean, `yamllint -c .yamllint` clean on the fragment.

**Not verified end to end.** I have no AD or LDAP user to test against - the lab array has one non-local admin but it already has array-side state. @meagan-gibbons-exp - if you are able to re-run your playbook against this branch it would be worth confirming before it ships.

Changelog fragment included: `changelogs/fragments/1061_purefa_user_ad_username.yaml`.


plugins/modules/purefa_user.py
```diff
@@ -279,31 +281,30 @@ def update_ad_user(module, array, user):
...
-        res = array.patch_admins(
-            names=[module.params["name"]],
-            admin=AdminPatch(public_key=module.params["public_key"]),
-        )
-        check_response(res, module, "Failed to add SSH key")
+    # An empty string clears the key, so this is checked against None rather
+    # than for truthiness. An AD user with no array-side state is not returned
+    # by get_admins, so there is no current key to compare against.
+    if module.params["public_key"] is not None:
+        current_key = getattr(user, "public_key", None) if user else None
```
meagan-gibbons-exp, line 299: This will never work. The array returns '****', not the actual key. I just drew up a very similar PR; see my changes. #1062

meagan-gibbons-exp:
I think we could use a merge of the two PRs. You caught check_mode I missed & your tests are probably better, but mine accounts for setting public keys not being able to be idempotent. Also, I added a separate regex for validating ad users without skipping input validation altogether

sdodsley:
Closing in favour of #1062, which now carries everything from here. Agreed on the merge — you were right on the substantive point.

**You were right about idempotency and I was wrong.** I checked against an array at REST 2.54: setting any key makes `get_admins` report `public_key` as `'****'`, whatever the key, and clearing it returns `None`.

```
key A set (97 chars) -> get_admins returns '****'
key B set (60 chars) -> get_admins returns '****'
cleared              -> get_admins returns None
```

So a key can never be compared against the one supplied, but its presence can be — which is exactly what your PR says, and why removal is idempotent while setting is not. My "idempotency fix" here was fictional: the comparison was really running against `'****'`, so it always patched anyway, and the test asserting otherwise mocked a state the array never produces. Dropped rather than carried.

**You were also right that skipping validation entirely was wrong**, and your test docstring gave the reason. It is not theoretical:

```
get_admins(names=['test-admin1', 'test-admin2'])  ->  200, 2 admins
get_admins(names=['test-admin1,test-admin2'])     ->  200, 2 admins   # ONE name string
```

One name containing a comma is sent, and acted on, as two. On `patch_admins` that means a single `name:` value modifies two admins.

**Two changes to the pattern itself**, pushed to your branch:

The array is happy with non-ASCII names — `get_admins(names=['josé.garcía'])` answers "Unable to find specified user", an existence error, the same answer it gives for a dotted name. An ASCII allowlist would reject directory users that work fine, which is the same class of bug this PR fixes. It now rejects only what genuinely cannot work: whitespace, and the characters that change which admins are targeted.

And `$` matches before a trailing newline in Python, so `^...$` accepted `"alice\n"` — which is what `lookup('file', 'username.txt')` gives you. Anchored with `\A` and `\Z` instead, with a test case for it.

**One thing your branch was still missing:** the API token block for an *existing local* user was unguarded for check mode, so `--check` still revoked and reissued a live token. You had guarded the SSH key beside it. The carried-over test is what caught it — that path is the more damaging of the two, since it invalidates a credential already in use.

**Also worth saying:** your comment says you missed `check_mode`, but your diff had the guards in both `create_local_user` and `update_ad_user` — you only missed the one token path above.

Both patterns are now module-level constants, so the tests reference them directly instead of scraping the module source; that keeps the drift guard without a test that passes even when the validation is never invoked.

Still worth you running your original playbook against #1062 before it ships — I have no AD user to test against, and that is the one thing none of this verifies.

PR #1074, opened 2026-09-21
purefa_info: add servers gather_subset

### LINKED ISSUE #1073: Add servers into gather_subset list of purefa_info module
**Is your feature request related to a problem? Please describe.**
Hello folks.

I noticed that purefa_server module was added recently. I would be nice to add "servers" in gather_subset parameter. I am unsure if the latest REST API version support this

Gather current file servers.

Thank you

ananthakrishnanvk:
##### SUMMARY

Adds `servers` to the `gather_subset` parameter of `purefa_info`. Each entry reports the file server's DNS config, directory service, local directory service, and the network interfaces attached to it. Gated on REST API 2.44. Arrays on older firmware skip the subset silently when `all` is requested.

Fixes #1073 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- purefa_info

##### ADDITIONAL INFORMATION

- `changelogs/fragments/1074_purefa_info_servers.yaml`.
(https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)

```
"servers": {
            "_server_1": {
                "directory_service": null,
                "dns": "file",
                "local_directory_service": "domain",
                "network_interfaces": [
                    "filevip"
                ]
            }
}
```


plugins/modules/purefa_info.py
```diff
@@ -3111,6 +3112,35 @@ def generate_tgroups_dict(array):
...
 
 
+def generate_servers_dict(array):
+    servers_info = {}
+    res = array.get_servers()
+    if res.status_code != 200:
+        return servers_info
+    servers = list(res.items)
+    # Build a map of interface name → list of server names it is attached to
+    iface_res = array.get_network_interfaces()
```
sdodsley: this call is made even if there are no servers, which is a wasted API call. `get_servers()` will return `{}` and a 200 with no servers configured. We might want to have a warning message saying "No servers configured", the same way `generate_config_dict()` warns if FA-Files is not enabled. This would require passing `module` to the function.

plugins/modules/purefa_info.py
```diff
@@ -3111,6 +3112,35 @@ def generate_tgroups_dict(array):
     return tgroups_info
 
 
+def generate_servers_dict(array):
+    servers_info = {}
+    res = array.get_servers()
+    if res.status_code != 200:
+        return servers_info
+    servers = list(res.items)
+    # Build a map of interface name → list of server names it is attached to
```
sdodsley: wrong way around - should be server names -> interfaces

sdodsley (changes requested):
Can you also add some unit tests for this new function?

plugins/modules/purefa_info.py
ananthakrishnanvk: Fixed.
The comment had the mapping direction backwards. The code was correct

plugins/modules/purefa_info.py
ananthakrishnanvk: Done.
`get_network_interfaces()` is guarded by `if servers:` so it only runs when there's at least one server. Also added `module.warn("No file servers are configured on this array")` on the empty-list path, matching the pattern you pointed to in `generate_config_dict`.

